PR #1, opened 2020-08-24
Add initial readme, license and CI

aghecenco:


.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
+.idea/
+Cargo.lock
```
andreeaflorescu, line 3: Since this will generate a binary, I don't think we want to exclude `Cargo.lock` anymore.

.gitignore
```diff
@@ -0,0 +1,3 @@
+/target
+.idea/
```
andreeaflorescu, line 2: `.idea\` should not be part of the gitingore of the individual repository, but rather can be defined on the devs machine as part of the global gitignore on the system level. The reason is that different people will use different IDEs (or no IDEs at all - hardcore), and gitignore would be polluted this way.

andreeaflorescu (commented):
High level it looks good to me. I think it would be interesting to add more details about the design (i.e. interfaces that we expect to have, high level components?). Are we going to add these details in a subsequent PR?

aghecenco:
> High level it looks good to me. I think it would be interesting to add more details about the design (i.e. interfaces that we expect to have, high level components?). Are we going to add these details in a subsequent PR?

Yup. I'd like to have the skeleton in place, and follow up with details (as they become apparent to me). It'd also be great to get a discussion going here on what folks expect from the reference VMM.

.gitignore
aghecenco: Removed

.gitignore
aghecenco: Updated

aghecenco:
@michael2012z WDYT about the (very high level) proposal in the readme?

michael2012z:
@aghecenco It's a good high level description. The purposes of the reference VMM are well defined.

lauralt (approved):

andreeaflorescu (approved):

alexandruag (approved):

PR #11, opened 2020-09-08
Reference VMM MVP

aghecenco:
This PR introduces the first iteration of a rust-vmm-based reference VMM.


README.md
```diff
@@ -1,41 +1,57 @@
...
-## Examples
+The reference VMM can be used out of the box as a `hello-world` example of a
+fully functional VMM built with `rust-vmm` crates.
+
+```bash
+cargo run                       \
+    --guest-memory=1024         \
+    --vcpus=1                   \
+    --kernel=/path/to/vmlinux   \
+    --cmdline="cmdline"
```
lauralt, line 37: This should be updated either here or in #1 to its current form.

src/vmm/src/vcpu/mod.rs
```diff
@@ -0,0 +1,264 @@
...
+                                .serial
+                                .lock()
+                                .unwrap()
+                                .0
+                                .write((addr - 0x3f8) as u8, data[0])
+                                .is_err()
+                            {
+                                eprintln!("Failed to write to serial port");
+                            }
+                        } else if 0x060 <= addr && addr < (0x060 + 5) {
```
lauralt, line 240: Aren't only 0x60, 0x61 and 0x64 - i8042 ports?

andreeaflorescu (dismissed):
We should not add binary blobs to GitHub:
- vmlinux-hello-busybox
- vmlinux-hello-busybox-halt

aghecenco:
> We should not add binary blobs to GitHub:
> 
>     * vmlinux-hello-busybox
> 
>     * vmlinux-hello-busybox-halt

Replaced with a post-checkout hook that build the 2nd image, which is used in the tests.
The 1st one is for demo purposes, and there's a script in resources/kernel for it.

src/vmm/src/vcpu/mod.rs
aghecenco: I actually have no idea :smiley: thanks for pointing it out! Is there a recommended link I could link here as "ground truth"?

jdub:
This is incredibly readable and straightforward! 🤘🏻

tests/test_run_reference_vmm.py
```diff
@@ -0,0 +1,46 @@
...
+        ),
+        "--vcpus", "num_vcpus={}".format(num_vcpus)
+    ]
+    
+    # We can't talk to the reference VMM yet, and it can't talk back.
+    # If we try to capture the output, Python doesn't return control to the
+    # test until the child process ends, which it doesn't. So we have to trust
+    # that the output is there, let the VMM run for a bit, then kill it.
+    # In the future, it will communicate via metrics / devices.
+    vmm_process = subprocess.Popen(vmm_cmd)
```
lauralt, line 41: Hmmm, are we sure that this test actually runs `vmm_cmd` successfully? `subprocess.Popen` seems to have a weird behavior (i.e. it fails, from what I noticed, only if the binary doesn't exist (in this case `cargo`, which exists)).
Wouldn't it be better to run `subprocess.run` for example or something else?

Also, funny question: is the `cargo run ...` option still valid with the workspace changes that I mentioned in another comment? I didn't manage to run it successfully now, but I might be doing something really wrong :-?.

src/vmm/src/vcpu/mod.rs
lauralt: This one seems trustworthy: https://web.archive.org/web/20200602023256/https://wiki.osdev.org/%228042%22_PS/2_Controller#PS.2F2_Controller_IO_Ports, but I couldn't find some "official" list of the IO ports meaning.

lauralt (commented):
First part of review.

I'm a bit intrigued by the `Cargo.toml` workspace changes and how they deal with `cargo build`/`cargo run` commands. Would really appreciate if someone has more info here.

docs/DESIGN.md
```diff
@@ -0,0 +1,242 @@
...
+and the VMM starts with them. To alter its structure, the VMM process must be
+killed and re-run with the modified parameters. The core of the reference VMM
+is the `vmm` crate, which encapsulates and binds together all the other
+functionality-providing crates; the user-facing component is minimalistic and
+meant to be discarded when the VMM is forked for new usage, in order for the
+user to provide an UI suitable to their needs.
+
+The first iteration of the reference VMM contains a subset of the crates
+available in `rust-vmm`: `kvm-ioctls`, `kvm-bindings`, `vm-memory`,
+`linux-loader`, `event-manager` and `vm-superio`. Only the serial console
```
lauralt, line 17: Should we mention here about `vmm-sys-util` as well?

lauralt (commented):
Second part of review.

tests/test_run_reference_vmm.py
aghecenco: Unlike `Popen`, `subprocess.run` doesn't end until the child process ends - so the test would be stuck in it.
As for the workspace stuff, I'm looking into those, it's not clear to me yet how we want to proceed.

lauralt: From what I manually tested, the things mentioned here: https://github.com/rust-vmm/rust-vmm-ci/issues/38 seem to fix the workspace issue.

src/cli/src/lib.rs
```diff
@@ -0,0 +1,254 @@
...
+            )
+            .arg(
+                Arg::with_name("kernel")
+                    .long("kernel")
+                    .required(true)
+                    .takes_value(true)
+                    .validator(Self::validate_kernel_config),
+            )
+            .get_matches_from_safe(cmdline_args)
+            .map_err(|_| "Failed to parse command line arguments".to_string())?;
```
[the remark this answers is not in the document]
aghecenco: I haven't figured out a way to do either. To me it seemed like the easiest way to validate those strings was to try and cast them into `*Config` objects, and consider them valid if the cast succeeds. As this CLI is meant to be for demonstrative purposes only, and throw-away, I didn't think too much of anything fancier :smile: The cleanest way IMO would be to replace `clap` like you mentioned in another comment, in the meantime I think we can compromise on the CLI at least until we commit to something as production-worthy as the VMM - WDYT?

src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+use vmm::VMM;
+
+fn main() {
+    #[cfg(target_arch = "x86_64")]
+    {
+        let vmm_config = CLI::launch(
+            env::args()
+                .collect::<Vec<String>>()
+                .iter()
+                .map(|s| s.as_str())
```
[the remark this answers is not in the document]
aghecenco: I started off with `String`s, then reverted to slices while writing the tests - everything had to be `.to_string()`'ed and `.clone()`'ed. It made the code overall more verbose and `clone`-y.

README.md
```diff
@@ -1,41 +1,79 @@
...
+## Usage
 
-```rust
-use my_crate;
+The reference VMM can be used out of the box as a `hello-world` example of a
+fully functional VMM built with `rust-vmm` crates.
 
-...
+```bash
+vmm-reference                                                           \
```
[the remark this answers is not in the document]
aghecenco: Added a doc as the `getting started` section might get fluffier once we start adding devices and extending platform support. Added the `cmdline` in the example too.
I reverted the workspace settings to the previous ones (empty `[workspace]`), but we'll have to fix rust-vmm-ci and update the hook before the tests pass again :smile:

lauralt (commented):
A few other comments. Still have to review the `vmm` crate.

docs/DESIGN.md
```diff
@@ -0,0 +1,243 @@
...
+Currently, with no available devices except for the serial console, the
+reference VMM can boot a simple Linux kernel with a baked-in `initramfs`.
+`*Config` structs passed by the UI are cached in a `VmmConfig` struct until
+`run()`, when the VMM is instantiated and the configurations are turned into
+live objects.
+
+The steps to running a guest, presuming the configurations are parsed and
+stored, are as follows:
+
+1. Set up KVM. This is done through `kvm-ioctls`. It creates the KVM virtual
```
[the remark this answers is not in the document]
aghecenco: Linked several lines above, where the list of used crates is mentioned

src/cli/src/lib.rs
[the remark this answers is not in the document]
aghecenco: You're absolutely right that the user experience is pretty crappy and needs more work, and thanks for the reminder of why we're doing this in the first place :smile: 
I'm working on nicer errors.
> call `get_matches_from()` instead of `get_matches_from_safe()`

The `_safe` one returns a testable `Result`. The other one `exit`s the process, so there's no way to unit test the CLI interaction. How about a compromise: keep the `_safe()` one and the tests, but print the usage in case of errors?

src/vmm/tests/integration_tests.rs
```diff
@@ -0,0 +1,82 @@
...
+fn default_memory_config() -> MemoryConfig {
+    MemoryConfig { mem_size_mib: 1024 }
+}
+
+fn default_kernel_config() -> KernelConfig {
+    let kernel_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+        .join("../../resources/kernel/vmlinux-hello-busybox-halt");
+    // Sanity check. Because the VMM runs in a separate process, if the file doesn't exist,
+    // all we see is a different exit code than 0.
+    assert!(kernel_path.as_path().exists());
```
lauralt, line 29: While reviewing https://github.com/rust-vmm/rust-vmm-ci/pull/37, I noticed that even if this path doesn't exist, so the assert fails/should fail, the test is still passing. I assume this is not the desired behavior. Can you expand a bit on how is this exit code propagated further?

resources/kernel/make_busybox.sh
```diff
@@ -0,0 +1,103 @@
...
+# Prepare busybox.
+echo "Downloading busybox..."
+mkdir -p busybox_rootfs
+curl "$BUSYBOX_URL" > "$BUSYBOX_ARCHIVE"
+echo "Extracting busybox..."
+tar xf "$BUSYBOX_ARCHIVE"
+pushd "$BUSYBOX" &>/dev/null
+# Build statically linked busybox.
+cp "$TEST_RESOURCE_DIR/busybox_static_config" .config
+make -j
```
andreeaflorescu, line 46: This command is freezing my laptop really bad. Same for the other `make -j` :laughing: 

Any chance we can make this configurable instead? I am not sure it is a good idea even for the CI to run with `-j` because we might have multiple agents on the same host and they'll have no available resources to run their jobs.

src/vmm/tests/integration_tests.rs
aghecenco: Great catch! The default panic handler exited with 0, so the parent process thought the child was on the happy path. I moved the `assert` before the `fork` to avoid the unnecessary new process on the error path, and installed a panic hook that exits a panicky process with a nonzero exit code.

resources/kernel/make_busybox.sh
aghecenco: Done. You can now optionally pass `-j <as many as you want>`. The script got ugly, so I refactored it a bit. PTAL

aghecenco:
@andreeaflorescu the changes you requested are dealt with, the binaries are out of the repo.

resources/kernel/make_busybox.sh
aghecenco: Tests take forever now, though. WDYT would be a sweet spot for num cpus to assign?

src/vmm/src/boot.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    // Add entries for the usable RAM regions (potentially surrounding the MMIO gap).
+    let last_addr = guest_memory.last_addr();
+    if last_addr < mmio_gap_start {
+        add_e820_entry(
+            &mut params,
+            himem_start.raw_value() as u64,
+            last_addr
+                .checked_offset_from(himem_start)
+                .ok_or(Error::HimemStartPastMemEnd)?,
```
lauralt, line 94: Hmmm, if `last_addr` is the maximum inclusive address, shouldn't the result of `checked_offset_from` be incremented with 1 to get the size of the region?

src/vmm/src/boot.rs
```diff
@@ -0,0 +1,121 @@
...
+            E820_RAM,
+        )?;
+    } else {
+        add_e820_entry(
+            &mut params,
+            himem_start.raw_value(),
+            mmio_gap_end
```
lauralt, line 101: Shouldn't this be `mmio_gap_start`?

src/vmm/src/config.rs
```diff
@@ -0,0 +1,160 @@
...
+    ParseVcpus(String),
+}
+
+impl fmt::Display for ConversionError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        use ConversionError::*;
+        match self {
+            ParseKernel(ref s) => write!(f, "Invalid input for memory: {}", s),
+            ParseMemory(ref s) => write!(f, "Invalid input for memory: {}", s),
+            ParseVcpus(ref s) => write!(f, "Invalid input for memory: {}", s),
```
lauralt, line 25: These should be updated with kernel/memory/vcpus.

src/vmm/src/boot.rs
aghecenco: That's right, nice catch! Fixed, and added comments for the unchecked additions.

src/vmm/src/boot.rs
aghecenco: Oops, yes it should. Looks like I should have added a test that boots with more memory.

src/vmm/src/config.rs
aghecenco: Done

lauralt (commented):
Added a few more comments (the review takes longer than expected :laughing:, sorry :( ).

src/vmm/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// * `vcpu_cfg` - [`VcpuConfig`](struct.VcpuConfig.html) struct containing vCPU configurations.
+    pub fn configure_vcpus(
+        &mut self,
+        vcpu_cfg: VcpuConfig,
+        kernel_load: KernelLoaderResult,
+    ) -> Result<()> {
+        mptable::setup_mptable(&self.guest_memory, vcpu_cfg.num_vcpus)
+            .map_err(|e| Error::Vcpu(vcpu::Error::Mptable(e)))?;
+
+        let base_cpuid = self
```
lauralt, line 314: Making this `mut` will no longer require having the `vcpu_cpuid` clone a few lines below.

src/vmm/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+
+    /// Configure guest vCPUs.
+    ///
+    /// # Arguments
+    ///
+    /// * `vcpu_cfg` - [`VcpuConfig`](struct.VcpuConfig.html) struct containing vCPU configurations.
+    pub fn configure_vcpus(
+        &mut self,
+        vcpu_cfg: VcpuConfig,
+        kernel_load: KernelLoaderResult,
```
lauralt, line 309: The arguments list should be updated with this one. Also, isn't  the `kernel_load` GuestAddress field from `KernelLoaderResult` enough here or do we expect to use in the future the other fields too?

src/vmm/src/lib.rs
```diff
@@ -0,0 +1,404 @@
...
+
+            // Configure regs, sregs and fpu.
+            vcpu.configure_regs(kernel_load.kernel_load)
+                .map_err(Error::Vcpu)?;
+            vcpu.configure_sregs(&self.guest_memory)
+                .map_err(Error::Vcpu)?;
+            vcpu.configure_fpu().map_err(Error::Vcpu)?;
+
+            // Configure LAPICs.
+            vcpu.configure_lapic().map_err(Error::Vcpu)?;
```
lauralt, line 351: I would move the vcpu configuration to a separate function. Might make things more readable.

tests/test_run_reference_vmm.py
lauralt: I checked again, the test is still passing if, for example, we replace `run` with `blah` in `vmm_cmd`. In this case, the test is not useful. I'll try looking more into how we can make this test fail if the command is invalid.

aghecenco: I think testing `cargo blah` instead of `cargo run` tests `cargo` more than it does the VMM :thinking:

src/vmm/src/boot.rs
aghecenco: I'll add an issue for that :arrow_up:

src/vmm/src/config.rs
```diff
@@ -0,0 +1,160 @@
...
+
+impl TryFrom<String> for VcpuConfig {
+    type Error = ConversionError;
+
+    fn try_from(vcpu_cfg_str: String) -> result::Result<Self, Self::Error> {
+        // Supported options: `num_vcpus=<u8>`
+        let vcpu_cfg_str_lower = vcpu_cfg_str.to_lowercase();
+        let tokens: Vec<&str> = vcpu_cfg_str_lower
+            .split('=')
+            .filter(|tok| !tok.is_empty())
```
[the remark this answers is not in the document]
aghecenco: Agreed that there's ample room for future improvements to the CLI :+1:

src/vmm/src/devices.rs
```diff
@@ -0,0 +1,80 @@
...
+                let event_set = events.event_set();
+                let unregister_condition =
+                    event_set.contains(EventSet::ERROR) | event_set.contains(EventSet::HANG_UP);
+                if count > 0 {
+                    if self.0.enqueue_raw_bytes(&out[..count]).is_err() {
+                        eprintln!("Failed to send bytes to the guest via serial input");
+                    }
+                } else if unregister_condition {
+                    // Got 0 bytes from serial input; is it a hang-up or error?
+                    ops.remove(events)
```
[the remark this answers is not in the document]
aghecenco: The subscriber (serial device) gets removed [here](https://github.com/rust-vmm/event-manager/blob/abac92899908e6a23a6927b5c1c9d4d53bc4f59e/src/events.rs#L267).

src/vmm/src/vcpu/mod.rs
```diff
@@ -0,0 +1,265 @@
...
+///
+/// This struct is a temporary (and quite terrible) placeholder until the
+/// [`vmm-vcpu`](https://github.com/rust-vmm/vmm-vcpu) crate is stabilized.
+pub(crate) struct Vcpu {
+    /// Index.
+    pub index: u8,
+    /// KVM file descriptor for a vCPU.
+    pub vcpu_fd: VcpuFd,
+    /// Shared device.
+    pub serial: SafeStdoutSerial,
```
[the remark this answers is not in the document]
aghecenco: This is an absolute hack. It's here while we figure out the device & device manager interfaces elsewhere. I considered adding a device manager but without stable interfaces it was starting to turn out into a lot of throw-away code for a very simply use case, so I took the quick way out.
The next thing to do - https://github.com/rust-vmm/vmm-reference/issues/13 - is circle back to `vm-device` and `vm-virtio` and change this hack into something respectable :smile:

tests/test_run_reference_vmm.py
lauralt: It was just an example, replacing `--memory` with some invalid arg will keep the test passing :(. In my opinion we should make sure that whatever `subprocess` function we choose, the test is passing only if `cargo run ...` (or even `cargo blah`) is successful.

src/vmm/src/lib.rs
aghecenco: I don't know if this is possible, but what if `filter_cpuid` changed something in one vCPU's CPUID entries, that isn't correct for the next vCPU?

andreeaflorescu: For now I think it is better to create a clone of `base_cpuid` because in the simple model that we are implementing here we start from the same CPUID (the one reported by KVM), and just selectively update the CPUID leafs that we are interested in.

A case where we need to override the CPUID entry to something that is not valid for all VCPUs is leaf 1 where we specify what is the id of the VCPU. Since this value will be overwritten by all VCPUs, it won't actually become invalid, but anyway.

src/vmm/src/devices.rs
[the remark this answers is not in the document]
lauralt: This is (only) removing the subscriber from the `EpollWrapper` watch list. I was referring to [this](https://github.com/rust-vmm/event-manager/blob/master/src/manager.rs#L39). Anyway I had a chat with @andreeaflorescu and indeed this has to be done explicitly only if the app continues the execution, but event monitoring is not necessary (which shouldn't be the case here).

docs/getting-started.md
```diff
@@ -0,0 +1,109 @@
...
+
+```wrap
+cargo run --                      \
+    --memory mem_size_mib=1024    \
+    --kernel path=${PWD}/resources/kernel/vmlinux-hello-busybox,cmdline="console=ttyS0 i8042.nokbd reboot=k panic=1 pci=off",himem_start=1048576  \
+    --vcpus num_vcpus=1
+```
+
+```wrap
+cargo build
```
andreeaflorescu, line 104: Should we use release instead?

docs/getting-started.md
```diff
@@ -0,0 +1,109 @@
...
+
+### Putting It All Together
+
+Once all the prerequisites are met, the reference VMM can be run either
+directly through `cargo`, passing on its specific
+[command line arguments](../README.md#cli-reference), or after building it with
+`cargo build`.
+
+```wrap
+cargo run --                      \
```
andreeaflorescu, line 97: Should we use release here as well?

src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+                .collect(),
+        )
+        .expect("Failed to parse command line options");
```
andreeaflorescu, line 27: Instead of this expect, can we do something like:
```rust
        if let Ok(vmm_config) = CLI::launch(
            env::args()
                .collect::<Vec<String>>()
                .iter()
                .map(|s| s.as_str())
                .collect(),
        ) {
            let mut vmm = VMM::try_from(vmm_config).expect("Failed to create VMM from configurations");
            vmm.run();
        }
    }
```

The problem with expect is that the program terminates with an ugly panic as follows if you don't pass all the required parameters.
```
thread 'main' panicked at 'Failed to parse command line options: "Invalid command line arguments: \u{1b}[1;31merror:\u{1b}[0m The following required arguments were not provided:\n    \u{1b}[1;31m--kernel <kernel>\u{1b}[0m\n    \u{1b}[1;31m--memory <memory>\u{1b}[0m\n    \u{1b}[1;31m--vcpus <vcpus>\u{1b}[0m\n\nUSAGE:\n    vmm-reference --kernel <kernel> --memory <memory> --vcpus <vcpus>\n\nFor more information try \u{1b}[32m--help\u{1b}[0m\n"', src/main.rs:27:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

src/vmm/src/lib.rs
```diff
@@ -0,0 +1,405 @@
...
+                };
+
+                // Safe because the fd is a valid KVM file descriptor.
+                unsafe { self.vm_fd.set_user_memory_region(memory_region) }
+            })
+            .map_err(Error::KvmIoctl)?;
+
+        self.vm_fd
+            .set_tss_address(KVM_TSS_ADDRESS as usize)
+            .map_err(Error::KvmIoctl)?;
```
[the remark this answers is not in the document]
aghecenco: From my understanding it's the other way around, but maybe I got it wrong - could you explain a bit why it's unnecessary?
Here's where I'm coming from - I know the guest boots without knowing the TSS addr but [this snippet](https://elixir.bootlin.com/linux/latest/source/arch/x86/kvm/vmx/vmx.c#L2858) when the guest enters real mode looks for it:
```c
static void enter_rmode(struct kvm_vcpu *vcpu)
{
     ...
	/*
	 * Very old userspace does not call KVM_SET_TSS_ADDR before entering
	 * vcpu. Warn the user that an update is overdue.
	 */
	if (!kvm_vmx->tss_addr)
		printk_once(KERN_WARNING "kvm: KVM_SET_TSS_ADDR need to be "
			     "called before entering vcpu\n");

    ...
	vmcs_write32(GUEST_TR_LIMIT, RMODE_TSS_SIZE - 1);
```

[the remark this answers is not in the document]
andreeaflorescu: If you look at the comment above, only "very old userspace does not call KVM_SET_TSS_ADDR" which means that for old hardware you need to manually call it. Also, looking at the `KVM_SET_TSS_ADDR` it seems that for unrestricted guests, this is a no-op: https://elixir.bootlin.com/linux/latest/source/arch/x86/kvm/vmx/vmx.c#L4630.

More about unrestricted guests in the Intel SDM chapter "25.6 UNRESTRICTED GUESTS": https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3c-part-3-manual.pdf

In a nutshell, we are running an "unrestricted guest" because we are setting both CR0.PE and CR0.PG to 1:
- [CR0.PE](https://github.com/aghecenco/vmm-reference/blob/67557522737f36b8e41f33194ee0945c3b810833/src/vmm/src/vcpu/mod.rs#L147)
- [CR0.PG](https://github.com/aghecenco/vmm-reference/blob/67557522737f36b8e41f33194ee0945c3b810833/src/vmm/src/vcpu/mod.rs#L176)

Because we are running unrestricted guests, there is no need to call SET_TSS. Does that make sense?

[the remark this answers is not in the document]
aghecenco: Makes total sense. Thank you for the detailed explanation :blue_book:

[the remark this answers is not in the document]
andreeaflorescu: So obviously something was wrong in my explanation:

> In a nutshell, we are running an "unrestricted guest" because we are setting both CR0.PE and CR0.PG to 1:

KVM cares not of what we set in `CR0.PE` and `CR0.PG`. Instead it decides if `enable_unrestricted_guest` is true by (eventually, in a really long chain of functions) calling `rdmsr` (which is short for read msr) on the host and checking if `SECONDARY_EXEC_UNRESTRICTED_GUEST` is 1.

That being now corrected, it still seems like unrestricted guests is a feature that was introduced first in Westemere (according to my friend [wikipedia](https://en.wikipedia.org/wiki/X86_virtualization#Intel_virtualization_(VT-x))), which was released in 2010, so I still think we should be reasonably safe in deleting `SET_TSS_ADDR` because we're not planning on running on older hardware.

src/vmm/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version = "0.1.0"
+authors = ["rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
+edition = "2018"
+
+[dependencies]
+event-manager = "0.2.0"
+kvm-bindings = { version = "0.3.0", features = ["fam-wrappers"] }
+kvm-ioctls = "0.5.0"
+linux-loader = "0.2.0"
+vm-device = { git = "https://github.com/aghecenco/vm-device.git", branch = "send" }
```
aghecenco, line 12: This is temporary - will point to upstream `vm-device` once https://github.com/rust-vmm/vm-device/pull/31 is merged, or some other way to ensure thread safe bus accesses makes its way there.

src/vmm/src/vcpu/mod.rs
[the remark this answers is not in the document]
aghecenco: Hack removed. Things are done nicely through the bus now.

resources/kernel/make_busybox.sh
```diff
@@ -0,0 +1,213 @@
...
+HALT=
+# Number of CPUs to use during the kernel build process.
+MAKEPROCS=1
+# Flag for optionally cleaning the workdir and recompiling the kernel.
+CLEAN=
+
+# Supported arguments:
+# * `-h`: build a guest that halts.
+# * `-j`: compile with `make -j` (on all available CPUs).
+#         TODO: pass the value for `-j`.
```
lauralt, line 34: Also passing the VMLINUX name as arg could be a good TODO. Looks like if we modify something in the script (and we don't want to recompile the kernel), we have to also change that name, otherwise the old image will be copied (or maybe we could overwrite de old image :-?).
Can we open an issue for future improvements to this script so we don't forget about them?

andreeaflorescu (approved):

resources/kernel/make_busybox.sh
```diff
@@ -0,0 +1,213 @@
...
+/bin/echo "                                                   "
+/bin/echo "                 _                                 "
+/bin/echo "  _ __ _   _ ___| |_    __   ___ __ ___  _ __ ___  "
+/bin/echo " | '__| | | / __| __|___\ \ / / '_ \ _ \| '_ \ _ \ "
+/bin/echo " | |  | |_| \__ \ ||_____\ V /| | | | | | | | | | |"
+/bin/echo " |_|   \__,_|___/\__|     \_/ |_| |_| |_|_| |_| |_|"
+/bin/echo "                                                   "
+/bin/echo "                                                   "
+/bin/echo "Hello, world, from the rust-vmm reference VMM!"
+EOF
```
lauralt, line 115:
```suggestion
mount -t sysfs none /sys
mknod -m 666 /dev/ttyS0 c 4 64
/bin/echo "                                                   "
/bin/echo "                 _                                 "
/bin/echo "  _ __ _   _ ___| |_    __   ___ __ ___  _ __ ___  "
/bin/echo " | '__| | | / __| __|___\ \ / / '_ \ _ \| '_ \ _ \ "
/bin/echo " | |  | |_| \__ \ ||_____\ V /| | | | | | | | | | |"
/bin/echo " |_|   \__,_|___/\__|     \_/ |_| |_| |_|_| |_| |_|"
/bin/echo "                                                   "
/bin/echo "                                                   "
/bin/echo "Hello, world, from the rust-vmm reference VMM!"
setsid  cttyhack sh
EOF
```

Using this get around, `/bin/sh: can't access tty; job control turned off` won't be printed anymore after the `Hello, world, ...`: https://stackoverflow.com/questions/36529881/qemu-bin-sh-cant-access-tty-job-control-turned-off. The other answer from here might be nicer, but I didn't manage to make it work yet.

lauralt (approved):
Looks good, just a few nits/leftovers, but they can be fixed in a subsequent PR.

PR #17, opened 2020-10-28
tests: improve test_reference_vmm

andreeaflorescu:
Run the tests with using the release build. Instead of doing a cargo
run, we now do it in 2 steps:
1. Build
2. Run

The reason is that with this approach, after step 2 we can just wait
for max 3 seconds for the process to exit. If it did exit, it means that
an error occurred. This wait enables us to print the error to stderr
instead of just failing the test after the test timeout expires.

### DIFF
--- a/tests/test_run_reference_vmm.py
+++ b/tests/test_run_reference_vmm.py
@@ -34,18 +34,26 @@ def test_reference_vmm():
     # vCPU config
     num_vcpus = 1
 
+    build_cmd = "cargo build --release"
+    subprocess.run(build_cmd, shell=True, check=True)
     vmm_cmd = [
-        "cargo", "run", "--",
+        "target/release/vmm-reference",
         "--memory", "mem_size_mib={}".format(mem_size_mib),
         "--kernel", "cmdline=\"{}\",path={},himem_start={}".format(
             cmdline, kernel_path, himem_start
         ),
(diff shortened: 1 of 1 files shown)


aghecenco (approved):

lauralt:
Not related to the changes in this PR, but are these [statements](https://github.com/rust-vmm/vmm-reference/blob/master/tests/test_run_reference_vmm.py#L51-L53) true? I think we are blocking in the `while` loop and we never actually get [here](https://github.com/rust-vmm/vmm-reference/blob/master/tests/test_run_reference_vmm.py#L63-L65) if we start a vmm that doesn't print the hello world message.
Shouldn't we add some timeout to that while? :-?

aghecenco:
The pipeline that runs this test times out after 5 minutes ([see here](https://github.com/rust-vmm/vmm-reference/blob/master/.buildkite/custom-pipeline.yml#L15)) so in the worst case scenario (boot hangs / someone alters the init script & the lines don't appear anymore) we won't keep the CI machine occupied forever. Even so, it'd be great to have a non-text-based method to identify a live guest, but that's a bigger discussion to have.

lauralt:
The problem I see here is if someone runs this test locally. The test hangs which is not great :(. Anyway, I will open an issue for this.

lauralt (approved):

PR #19, opened 2020-11-03
make_busybox: fix 'can't access tty; job control turned off'

lauralt:
more details in the last commit message.

### DIFF
--- a/resources/kernel/make_busybox.sh
+++ b/resources/kernel/make_busybox.sh
@@ -185,15 +185,7 @@ else
     VMLINUX="vmlinux-hello-busybox"
 fi
 
-# Step 2.a): we have it cached locally.
-if [ -f "$WORKDIR/$KERNEL/$VMLINUX" ]; then
-    echo "Found $VMLINUX in $WORKDIR/$KERNEL."
-    cp "$WORKDIR/$KERNEL/$VMLINUX" "$TEST_RESOURCE_DIR"
-    echo "Copied to $TEST_RESOURCE_DIR."
-    exit 0
-fi
-
-# Step 2.b): start from scratch.
+# Step 2: start from scratch.
 mkdir -p "$WORKDIR" && cd "$WORKDIR"
 
 # Step 3: acquire kernel sources & config.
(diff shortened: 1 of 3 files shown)


resources/kernel/make_busybox.sh
```diff
@@ -115,7 +115,7 @@ mount -t sysfs none /sys
 EOF
 
     if [ -z "$halt" ]; then
-        echo "exec /bin/sh" >>init
+        echo "setsid sh -c 'exec sh </dev/ttyS0 >/dev/ttyS0 2>&1'" >>init
```
aghecenco, line 118: I think this needs the full path to `sh`, otherwise this happens:
```
Hello, world, from the rust-vmm reference VMM!
setsid: can't execute 'sh': No such file or directory
[    0.737487] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00007f00
```

```suggestion
        echo "setsid /bin/sh -c 'exec /bin/sh </dev/ttyS0 >/dev/ttyS0 2>&1'" >>init
```

aghecenco: CI should fail because of that, but it doesn't actually rebuild the images because of this: https://github.com/rust-vmm/vmm-reference/blob/3f8a44d5ddb1824804a67abb947e0de66d47cac3/resources/kernel/make_busybox.sh#L187
We should figure out a way to force a kernel rebuild when necessary, without making a CI run last for hours.

lauralt: I removed this step, if we don't change something in the script that is related to kernel, the rebuild should not be retriggered.

aghecenco: I still see the kernel panic because of `/bin/sh` on a fresh build.

lauralt: It is rather strange as this happens neither in the CI nor on my machine :(. Anyway, I updated with the full path, thanks!

aghecenco (approved):

andreeaflorescu (approved):

Issue #20, opened 2020-11-04
Boot bzImage kernels


aghecenco:
`linux-loader` supports ELF and bzImage formats on x86_64, therefore so should the reference VMM.

PR #21, opened 2020-11-04
bzImage kernel format support

aghecenco:
#20

This PR introduces:
* bzImage changes in the reference VMM: the VMM will always try to load an ELF first, but if the kernel file is not an ELF, it will give bzImage a shot. Boot parameters and rip value are adjusted accordingly.
* revamp of the kernel image builder script: made it more modular, more parameterizable, and capable of baking the busybox initramfs in an ELF as well as a bzImage.
* integration test for the above script: it got complicated, so it kind of needed a test itself. However, the test takes a long time to run because it compiles kernels from scratch for multiple valid parameter combinations, so it's not run by the CI. It will be included in a long running tests battery instead.

### DIFF
--- a/resources/kernel/make_kernel.sh
+++ b/resources/kernel/make_kernel.sh
@@ -1,0 +1,185 @@
+#!/bin/bash
+
+# Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+# This script contains functions for compiling a Linux kernel and initramfs.
+
+KERNEL_URL_BASE="https://cdn.kernel.org/pub/linux/kernel"
+
+die() {
+    echo "[ERROR] $1" >&2
+    echo "$USAGE" >&2 # To be filled by the caller.
+    # Kill the caller.
+    if [ -n "$TOP_PID" ]; then kill -s TERM "$TOP_PID"; else exit 1; fi
+}
+
+pushd_quiet() {
+    pushd "$1" &>/dev/null || die "Failed to enter $1."
+}
+
+popd_quiet() {
+    popd &>/dev/null || die "Failed to return to previous directory."
+}
+
+# Usage:
+#   extract_kernel_srcs kernel_version
+extract_kernel_srcs() {
+    kernel_version="$1"
+
+    [ -z "$kernel_version" ] && die "Kernel version not specified."
+    
+    # This magic trick gets the major component of the version number.
+    kernel_major="${kernel_version%%.*}"
+    kernel_archive="linux-$kernel_version.tar.xz"
+    kernel_url="$KERNEL_URL_BASE/v$kernel_major.x/$kernel_archive"
+
+    echo "Starting kernel build."
+    # Download kernel sources.
+    echo "Downloading kernel..."
+    [ -f "$kernel_archive" ] || curl "$kernel_url" > "$kernel_archive"
+    echo "Extracting kernel sources..."
+    tar --skip-old-files -xf "$kernel_archive"
+}
+
+# Usage:
+#   make_kernel_config /path/to/source/config /path/to/kernel
+make_kernel_config() {
+    # Copy base kernel config.
+    # Add any custom config options, if necessary (currently N/A).
+    kernel_config="$1"
+    kernel_dir="$2"
+
+    [ -z "$kernel_config" ] && die "Kernel config file not specified."
+    [ ! -f "$kernel_config" ] && die "Kernel config file not found."
+    [ -z "$kernel_dir" ] && die "Kernel directory not specified."
+    [ ! -d "$kernel_dir" ] && die "Kernel directory not found."
+
+    echo "Copying kernel config..."
+    cp "$kernel_config" "$kernel_dir/.config"
+}
+
+# Usage:
+#   make_initramfs              \
+#       /path/to/kernel/dir     \
+#       /path/to/busybox/rootfs \
+#       [halt_value]
+make_initramfs() {
+    kernel_dir="$1"
+    busybox_rootfs="$2"
+    halt="$3"
+
+    [ -z "$kernel_dir" ] && die "Kernel directory not specified."
+    [ ! -d "$kernel_dir" ] && die "Kernel directory not found."
+    [ -z "$busybox_rootfs" ] && die "Busybox rootfs not specified."
+    [ ! -d "$busybox_rootfs" ] && die "Busybox rootfs directory not found."
+
+    # Move to the directory with the kernel sources.
+    pushd_quiet "$kernel_dir"
+
+    # Prepare initramfs directory.
+    mkdir -p initramfs/{bin,dev,etc,home,mnt,proc,sys,usr}
+    # Copy busybox.
+    echo "Copying busybox to the initramfs directory..."
+    cp -r "$busybox_rootfs"/* initramfs/
+
+    # Make a block device and a console.
+    pushd_quiet initramfs/dev
+    echo "Creating device nodes..."
+    rm -f sda && mknod sda b 8 0
+    rm -f console && mknod console c 5 1
+    rm -f ttyS0 && mknod ttyS0 c 4 64
+
+    make_init "$halt"
+
+    chmod +x init
+    fakeroot chown root init
+
+    # Pack it up...
+    echo "Packing initramfs.cpio..."
+    find . | cpio -H newc -o > ../initramfs.cpio
+    fakeroot chown root ../initramfs.cpio
+
+    # Return to kernel srcdir.
+    popd_quiet
+    # Return to previous directory.
+    popd_quiet
+}
+
+# Usage: validate_kernel_format format
+# Prints the lowercase format name, if one of "elf" or "bzimage".
+# Exits with error if any other format is specified.
+validate_kernel_format() {
+    format="$1"
+
+    kernel_fmt=$(echo "$format" | tr '[:upper:]' '[:lower:]')
+    if [ "$kernel_fmt" != "elf" ] && [ "$kernel_fmt" != "bzimage" ]; then
+        die "Invalid kernel binary format: $kernel_fmt."
+    fi
+    echo "$kernel_fmt"
+}
+
+# Usage: kernel_target format
+# Prints the `make` target that builds a kernel of the specified format.
+kernel_target() {
+    format=$(validate_kernel_format "$1")
+
+    case "$format" in
+    elf)        echo "vmlinux"
+                ;;
+    bzimage)    # This is the default target.
+                ;;
+    esac
+}
+
+# Usage: kernel_binary format
+# Prints the name of the generated kernel binary.
+kernel_binary() {
+    format=$(validate_kernel_format "$1")
+    
+    case "$format" in
+    elf)        echo "vmlinux"
+                ;;
+    bzimage)    echo "arch/x86/boot/bzImage"
+                ;;
+    esac
+}
+
+# Usage:
+#   make_kernel
+#       /path/to/kernel/dir \
+#       format              \
+#       [num_cpus_build]    \
+#       [/path/to/kernel/destination]
+make_kernel() {
+    kernel_dir="$1"
+    format="$2"
+    nprocs="$3"
+    dst="$4"
+
+    [ -z "$kernel_dir" ] && die "Kernel directory not specified."
+    [ ! -d "$kernel_dir" ] && die "Kernel directory not found."
+    [ -z "$format" ] && die "Kernel format not specified."
+    [ -z "$nprocs" ] && nprocs=1
+   
+    target=$(kernel_target "$format")
+    kernel_binary=$(kernel_binary "$format")
+
+    # Move to the directory with the kernel sources.
+    pushd_quiet "$kernel_dir"
+
+    # Build kernel.
+    echo "Building kernel..."
+    # Missing the quotes for $target is intentional: if the target is empty,
+    # quotes will cause it to be passed as an empty string, which `make`
+    # doesn't understand. No quotes => no empty string.
+    make -j "$nprocs" $target
+
+    if [ -n "$dst" ] && [ "$kernel_binary" != "$dst" ]; then
+        # Copy to destination.
+        cp "$kernel_binary" "$dst"
+    fi
+
+    # Return to previous directory.
+    popd_quiet
+}
--- a/src/vmm/src/boot.rs
+++ b/src/vmm/src/boot.rs
@@ -3,7 +3,7 @@
 
 use std::result;
 
-use linux_loader::bootparam::boot_params;
+use linux_loader::{bootparam::boot_params, loader::KernelLoaderResult};
 use vm_memory::{Address, GuestAddress, GuestMemory, GuestMemoryMmap};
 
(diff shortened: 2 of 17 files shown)


src/vmm/src/lib.rs
```diff
@@ -214,17 +219,29 @@ impl VMM {
...
+        let kernel_load = match Elf::load(
             &self.guest_memory,
             None,
             &mut kernel_image,
             Some(GuestAddress(kernel_cfg.himem_start)),
-        )
-        .map_err(Error::KernelLoad)?;
+        ) {
+            Ok(result) => result,
+            Err(loader::Error::Elf(elf::Error::InvalidElfMagicNumber)) => BzImage::load(
```
andreeaflorescu, line 229: I think it would be nicer to get the file type and then either parse it as elf or as bzimage. It looks like calling the `file` command gives you the information about it being elf/bzimage, which makes me think that there should be some helpers available in rust as well that we can use.

andreeaflorescu (commented):
Review still in progress.

src/vmm/src/lib.rs
aghecenco: I'm looking into this. Funnily enough, the only discernible way I can find is trial and error. ELF binaries have a magic number in the header (that's what `readelf` checks, `file` too) that `bzImage` files don't - which is obvious enough, as `bzImage` is a kernel-only format, while ELF is pretty much any linked and executable binary, as the name suggests. So, in order to discern an ELF kernel from a `bzImage`, the only thing we can do is:
* read the first 16 bytes and look for the ELF magic number ([source](https://man7.org/linux/man-pages/man5/elf.5.html))
* if it's there, it certainly isn't a `bzImage`, so we can carry on trying to boot it and if it was just an ordinary executable, we'll fail
* if it's missing, follow the kernel format ([source](https://www.kernel.org/doc/Documentation/x86/boot.txt)) and since we know for sure it's not ELF, use the `HdrS` field to discern `bzImage` from `something else`

This is...exactly what we do now, with the difference that we chain the attempts instead of packing them in a separate function.

andreeaflorescu: I also took a look at how `file` is able to decide what is the file format when it is printing it. It is using [`magic`](https://linux.die.net/man/5/magic), which looks like reading the header for both elf & bzimage and compare it with a magic header database. This would mean parsing the header twice, and since the parser is already doing that, I suppose we can just leave it as is. Sorry for derailing you.

resources/kernel/make_kernel.sh
```diff
@@ -0,0 +1,185 @@
...
+    pushd_quiet "$kernel_dir"
+
+    # Build kernel.
+    echo "Building kernel..."
+    # Missing the quotes for $target is intentional: if the target is empty,
+    # quotes will cause it to be passed as an empty string, which `make`
+    # doesn't understand. No quotes => no empty string.
+    make -j "$nprocs" $target
+
+    if [ -n "$dst" ]; then
```
lauralt, line 178: We should do the `cp` only if "$kernel_binary" and "$dst" are also different, otherwise if "$dst" is "vmlinux" for example, the `cp` will give the following error: `cp: 'vmlinux' and 'vmlinux' are the same file` (this happens if the user passes `vmlinux` as value to `-k` or doesn't provide `-k` at all).

lauralt (commented):
Review still in progress. I left only one comment.

resources/kernel/make_kernel.sh
lauralt: Also, it looks like the way we are using make_kernel(), `"$dst"` is always present. Would it make sense to remove [this line](https://github.com/rust-vmm/vmm-reference/pull/21/commits/793e04ea5f0e438a8f20593402e3bc4c0fdf2569#diff-26cd4e8b4eaa05f96c140feb26fb78afe7f0c4c21b8cfc526c4b90013459c231R86)? This way we would `cp` only if user provided a kernel image name.

src/vmm/src/vcpu/mod.rs
```diff
@@ -94,15 +99,36 @@ impl Vcpu {
     }
 
     /// Configure regs.
-    pub fn configure_regs(&self, kernel_load: GuestAddress) -> Result<()> {
+    pub fn configure_regs(
```
lauralt, line 102: Is it possible to test this one?

src/vmm/src/boot.rs
```diff
@@ -73,6 +73,7 @@ fn add_e820_entry(
 /// * `mmio_gap_end` - address where the MMIO gap ends.
 pub fn build_bootparams(
     guest_memory: &GuestMemoryMmap,
+    kernel_load: &KernelLoaderResult,
```
lauralt, line 76: The list of arguments should be updated with this one.

lauralt: And the description too I think.

src/vmm/tests/integration_tests.rs
```diff
@@ -50,13 +47,52 @@ fn wait_vmm_child_process(vmm_pid: i32) {
...
+            // Parent process: give the VMM some time, then check the exit code.
+            // It should terminate and return 0.
+            sleep(Duration::from_secs(5));
+            wait_vmm_child_process(vmm_pid);
+        }
+    }
+}
+
+#[test]
+fn test_dummy_vmm_bzimage() {
```
lauralt, line 82: Just curious, is it possible/makes any sense to have rust negative integ tests (in the future)?

tests/test_build_deps.py
```diff
@@ -0,0 +1,89 @@
...
+        )
+    
+    # Generate a kernel that halts, if specified. The script will append the
+    # "-halt" suffix.
+    if hlt:
+        script_cmd.extend(["-h"])
+        expected_image_path = "{}-halt".format(expected_image_path)
+    
+    return script_cmd, expected_image_path
+
```
lauralt, line 89: nit: looks like there are some spaces here that should be removed.

andreeaflorescu (commented):
I think we should also test this functionality end to end by updating the existing python integration test and do the same checks for bzimage as well.

Besides the already existing comments, lgtm.

aghecenco:
> I think we should also test this functionality end to end by updating the existing python integration test and do the same checks for bzimage as well.

It's already updated, the existing test is parameterized with the kernel path and it runs once with ELF and once with bzImage.

resources/kernel/make_kernel.sh
aghecenco: `KERNEL_BINARY` has to be *something* for the line immediately after that, which appends the `-halt` suffix if the user wants a halt-y kernel. I'll rework the `cp`, good catch on the error!

src/vmm/src/boot.rs
```diff
@@ -83,10 +84,16 @@ pub fn build_bootparams(
...
+    if let Some(hdr) = kernel_load.setup_header {
+        params.hdr = hdr;
+    } else {
+        params.hdr.boot_flag = KERNEL_BOOT_FLAG_MAGIC;
+        params.hdr.header = KERNEL_HDR_MAGIC;
+        params.hdr.kernel_alignment = KERNEL_MIN_ALIGNMENT_BYTES;
+    }
+    if params.hdr.type_of_loader == 0 {
+        params.hdr.type_of_loader = KERNEL_LOADER_OTHER;
+    }
```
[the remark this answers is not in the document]
aghecenco: Yup, the comment still stands, `KERNEL_LOADER_OTHER` is the "default" loader type selected unless we expect a particular known bootloader to take over (see `type_of_loader` section in [here](https://www.kernel.org/doc/html/latest/x86/boot.html?highlight=boot#details-of-header-fields)).
The check is useful in cases in which `type_of_loader` wasn't set in the `bzImage` file we loaded previously; if it was set to something, it was copied as-is (a few lines above), marking that a certain bootloader is expected (or not, if it was `0xFF`). If it remains unset, the guest will eventually complain (see https://github.com/cloud-hypervisor/cloud-hypervisor/issues/918). I'll add a comment.

src/vmm/tests/integration_tests.rs
aghecenco: I don't see why it wouldn't make sense. It's definitely possible, we can attempt to boot a VMM with invalid parameters (and skip the valid params check, hehe).

src/vmm/src/vcpu/mod.rs
aghecenco: It's covered by the integration tests, are you thinking unit tests for each function here?

lauralt: I was thinking of something similar to [this](https://github.com/firecracker-microvm/firecracker/blob/master/src/arch/src/x86_64/regs.rs#L263) (and insisting on the bzImage scenario).

src/vmm/src/boot.rs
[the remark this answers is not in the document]
lauralt: Thanks! The `which we aren't` part didn't seem accurate anymore. Can't we use a pre-registered boot-loader for bzImage?

resources/kernel/make_kernel.sh
lauralt: Hmm, can we add a mention about the suffix appending to [this description](https://github.com/rust-vmm/vmm-reference/pull/21/files#diff-26cd4e8b4eaa05f96c140feb26fb78afe7f0c4c21b8cfc526c4b90013459c231R47)? I would expect the resulting kernel image to have the exact name that I am passing to `-k` option.

src/vmm/tests/integration_tests.rs
```diff
@@ -50,10 +49,50 @@ fn wait_vmm_child_process(vmm_pid: i32) {
...
+        PathBuf::from(env!("CARGO_MANIFEST_DIR"))
+            .join("../../resources/kernel/bzimage-hello-busybox-halt"),
+    ));
+    // Sanity check. Because the VMM runs in a separate process, if the file doesn't exist,
+    // all we see is a different exit code than 0.
+    assert!(kernel_config.path.as_path().exists());
+
     let vmm_config = VMMConfig {
-        kernel_config: default_kernel_config(),
+        kernel_config: kernel_config,
```
andreeaflorescu, line 95: nit: we can just write `kernel_config` here.

src/vmm/tests/integration_tests.rs
```diff
@@ -22,12 +22,11 @@ fn default_memory_config() -> MemoryConfig {
     MemoryConfig { mem_size_mib: 1024 }
 }
 
-fn default_kernel_config() -> KernelConfig {
-    let kernel_path = PathBuf::from(env!("CARGO_MANIFEST_DIR"))
-        .join("../../resources/kernel/vmlinux-hello-busybox-halt");
-    // Sanity check. Because the VMM runs in a separate process, if the file doesn't exist,
-    // all we see is a different exit code than 0.
-    assert!(kernel_path.as_path().exists());
+fn default_kernel_config(path: Option<PathBuf>) -> KernelConfig {
```
andreeaflorescu, line 25: How about instead of having `path` as an option, we pass a parameter something like image_type for which we can add a define:
```rust
enum ImageType {
    BzImage,
    Elf,
}
```

Then, we could have something like:
```rust
let kernel_config = default_kernel_config(ImageType::Elf);
// or
let kernel_config = default_kernel_config(ImageType::BzImage);
```

src/vmm/tests/integration_tests.rs
```diff
@@ -50,10 +49,50 @@ fn wait_vmm_child_process(vmm_pid: i32) {
...
+    // all we see is a different exit code than 0.
+    assert!(vmm_config.kernel_config.path.as_path().exists());
+    match pid {
+        0 => {
+            set_panic_hook();
+            match VMM::try_from(vmm_config) {
+                Ok(mut vmm) => {
+                    vmm.run();
+                    // Shouldn't get here with this guest image. It will loop forever.
+                    unsafe { _exit(-1) };
```
andreeaflorescu, line 69: Is this still accurate? It looks like we are running with the halt image which means that it does not end here because of the way we are handling VcpuExit::Shutdown and VcpuExit::Hlt (i.e. with libc::exit).

I think it might make sense to actually just normally terminate the execution instead of using `libc::exit`. It would also make the testing simpler because we won't need to call fork. This last part is not in the scope of this PR, and we should handle it separately. But I think we should update the comment for now.

src/vmm/tests/integration_tests.rs
```diff
@@ -50,10 +49,50 @@ fn wait_vmm_child_process(vmm_pid: i32) {
 }
 
 #[test]
-fn test_dummy_vmm() {
+fn test_dummy_vmm_elf() {
```
andreeaflorescu, line 52: This test is very similar to `test_dummy_vmm_bzimage`. Can you create a helper function something like `run_vmm` that does most of the things besides creating the `vmm_config`, and then have the two tests calling into the helper function?

src/vmm/src/lib.rs
```diff
@@ -210,17 +215,29 @@ impl VMM {
...
+                &self.guest_memory,
+                None,
+                &mut kernel_image,
+                Some(GuestAddress(kernel_cfg.himem_start)),
+            )
+            .map_err(Error::KernelLoad)?,
+            Err(e) => {
+                return Err(Error::KernelLoad(e));
```
andreeaflorescu, line 233: It looks fairly easy to test this in the integration test by creating an empty `Tempfile` and then check that Vmm::try_from returns the KernelLoad error. Can you also add a test for this scenario?

src/vmm/tests/integration_tests.rs
aghecenco: I prefer having an optional `Path[Buf]` because in the future we'll likely want to run multiple kernels of the same format, so having a default one for each type of format would mean wed'd *still* have to overwrite the path for a non-default case.

src/vmm/tests/integration_tests.rs
aghecenco: Yes. If it ever gets to `exit(-1)`, something went wrong, because it's supposed to handle the `VcpuExit` by `exit(0)` earlier.

src/vmm/src/boot.rs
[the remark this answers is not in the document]
aghecenco: Nope. Bootloaders live on a device (typically the hard disk or something like a USB or optical drive) so until the VMM gets block devices, we don't have a lot of options.

src/vmm/tests/integration_tests.rs
andreeaflorescu: If we do have that in the future, I guess we can then have the `path` parameter as well. Right now it feels like an early optimization that is causing the path parameter to select the image type instead of only selecting the path of the binary. Alternatively we can just define `DEFAULT_ELF` and `DEFAULT_BZIMAGE` and always pass the path so that they're uniformly used.

src/vmm/tests/integration_tests.rs
andreeaflorescu: I was commenting more on the "It will loop forever" part which seems inaccurate.

andreeaflorescu (approved):

src/vmm/src/boot.rs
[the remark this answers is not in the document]
lauralt: Makes sense, thanks!

lauralt (commented):
Looks good!

lauralt (approved):

andreeaflorescu (approved):

PR #32, opened 2020-11-11
[RFC] Block device proposal

alexandruag:
This PR proposes a block device implementation based on abstraction from upstream (merged or under review) and some local logic as well (parts of which can become reusable at some point themselves). The block device is not properly tied to the command line yet; for now it's just hardcoded in the Vmm, which tries to open a file named "disk.ext4" from the current working directory when being run. There are some comments in the last commit regarding the order in which we create the irqchip relative to the devices; would be great to see what kind of restriction we're dealing with there (if any).

The request parsing and execution part is taken from Firecracker, until we can reuse something from rust-vmm. The next steps here are to see what kind of user interface changes we want to add to the application for block devices, and to start creating more tests if things look reasonable.


src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,168 @@
...
+            VIRTIO_F_VERSION_1 | VIRTIO_F_IN_ORDER | VIRTIO_F_RING_EVENT_IDX | VIRTIO_BLK_F_FLUSH;
+
+        // A block device has a single queue.
+        let queues = vec![Queue::new(args.mem, QUEUE_MAX_SIZE)];
+        let config_space = build_config_space(&args.file_path)?;
+        let virtio_cfg = VirtioConfig::new(device_features, queues, config_space);
+
+        let irqfd = EventFd::new(EFD_NONBLOCK).map_err(Error::EventFd)?;
+        args.vm_fd
+            .register_irqfd(&irqfd, args.mmio_cfg.gsi)
```
aghecenco, line 52: In the spirit of #14 I'd like to see this done elsewhere (`run()`?). The *empty `Block`* can hang around inert without a `VmFd` until it *really* needs to become a live device.

src/devices/src/virtio/block/request.rs
```diff
@@ -0,0 +1,362 @@
...
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub enum RequestType {
+    In,
+    Out,
+    Flush,
+    GetDeviceID,
+    Unsupported(u32),
+}
+
+impl From<u32> for RequestType {
```
aghecenco, line 175: I think it'd make more sense to couple this RFC with the other block device-related RFC https://github.com/rust-vmm/vm-virtio/pull/29 and use the abstractions from there.

src/vmm/src/lib.rs
```diff
@@ -416,6 +466,13 @@ impl TryFrom<VMMConfig> for VMM {
     fn try_from(config: VMMConfig) -> Result<Self> {
         let mut vmm = VMM::new()?;
         vmm.configure_guest_memory(config.memory_config)?;
+        // Moving irqchip creation earlier here, because we use the vm_fd to register irqfds
+        // for virtio devices as part of their construction, and the register_irqfd operation
+        // fails if the irqchip hasn't been created (or so it seems).
```
andreeaflorescu, line 471: The GIC on aarch64 can only be created after the vcpus are created. With that in mind, instead of having `vmm.temp_add_devs` here, you should have it after `vmm.configure_vcpus` which will make it work on both aarch64 & x86_64 transparently.

src/devices/src/virtio/block/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+        .map_err(Error::Seek)?;
+    // If the file size is actually not a multiple of sector size, then data at the very end
+    // will be ignored.
+    let num_sectors = file_size >> SECTOR_SHIFT;
+    // This has to be in little endian btw.
+    Ok(num_sectors.to_be_bytes().to_vec())
+}
+
+// Arguments required when building a block device.
+// TODO: Add read-only operation support as a quick next step.
```
lauralt, line 67: In the future we might want to have some sort of array for the supported operations since there are a few more other than `read-only`.

src/devices/src/virtio/block/request.rs
lauralt: It could be a good moment to do this, indeed. Do we plan to do this in the same PR or another one?

src/devices/src/virtio/block/simple_handler.rs
```diff
@@ -0,0 +1,99 @@
...
+    pub queue: Queue<M>,
+    pub disk: DiskProperties,
+}
+
+impl<M: GuestAddressSpace, S: SignalUsedQueue> SimpleHandler<M, S> {
+    fn process_chain(&mut self, mut chain: DescriptorChain<M>) -> result::Result<(), Error> {
+        match Request::parse(&mut chain) {
+            Ok(request) => {
+                let len;
+                // TODO: The executor could actually write the status itself, right?
```
lauralt, line 46: It should, still trying to figure out how to do this in a nice way :D.

lauralt (commented):
High level looks good, left a few comments for now.

src/devices/src/virtio/block/device.rs
alexandruag: I would really like to try out this approach because it helps delegate all the setup logic to the device, as long as the necessary "resources" (such as the `VmFd` for this particular implementation) are provided by the environment.

src/devices/src/virtio/block/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+fn build_config_space<P: AsRef<Path>>(path: P) -> Result<Vec<u8>> {
+    let file_size = File::open(path)
+        .map_err(Error::OpenFile)?
+        .seek(SeekFrom::End(0))
+        .map_err(Error::Seek)?;
+    // If the file size is actually not a multiple of sector size, then data at the very end
+    // will be ignored.
+    let num_sectors = file_size >> SECTOR_SHIFT;
+    // This has to be in little endian btw.
+    Ok(num_sectors.to_be_bytes().to_vec())
```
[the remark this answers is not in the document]
alexandruag: definitely! my bad

src/devices/src/virtio/block/mod.rs
alexandruag: Yes, that makes sense; we should def keep it in mind when adding all that stuff.

src/devices/src/virtio/block/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+// Arguments required when building a block device.
+// TODO: Add read-only operation support as a quick next step.
+pub struct BlockArgs<M> {
+    pub mem: M,
+    // The device uses this to create and register the queue handler at activation time.
+    pub endpoint: RemoteEndpoint<Box<dyn MutEventSubscriber + Send>>,
+    // Used to register irfqds and ioeventfds by the device.
+    pub vm_fd: Arc<VmFd>,
+    pub mmio_cfg: MmioConfig,
+    pub file_path: String,
```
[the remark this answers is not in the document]
alexandruag: sure

src/devices/src/virtio/block/request.rs
alexandruag: I've done this (by pulling the changes from your PR in the `vm-virtio` branch used as a depenency right now) in #41.

src/devices/src/virtio/block/simple_handler.rs
alexandruag: That's cool; we can def live with this as is for the time being.

src/vmm/src/lib.rs
alexandruag: I've pulled the changes you made in this direction for #41. Thanks!

alexandruag (commented):
Thanks for the comments! I've added answers here, and addressed them as part of #41 (which is meant to supersede this RFC PR). We can close this after that gets resolved/merged.

alexandruag:
Closing as the discussion moved over to #41.

PR #35, opened 2020-11-16
Create KvmVcpu and KvmVm abstractions

andreeaflorescu:
The long term view is that `KvmVcpu` and `KvmVm` become implementations of hypervisor agnostic `Vm` and `Vcpu` abstractions.

For now, this patch encapsulates the Vm & Vcpu functionality, and exports a simpler interface to the VMM, with the goal of making it as easy as possible to use the `vm-vcpu` crate. The configuration of various registers now happens behind the scene when `KvmVcpu` is initialized.


src/vmm/src/lib.rs
```diff
@@ -361,11 +345,9 @@ impl VMM {
             eprintln!("Failed to set raw mode on terminal. Stdin will echo.");
         }
 
-        for mut vcpu in self.vcpus.drain(..) {
-            let _ = thread::Builder::new().spawn(move || loop {
-                vcpu.run();
-            });
-        }
+        // TODO: should we handle this in another way rather than a panic?
```
aghecenco, line 307: Yep, I think so. The `vmm` crate here is a component in itself and it should be up to the app that consumes it to deal with a VMM that won't start. Maybe the error will be recoverable at some point and the UI could offer a retry possibility.

src/vmm/src/vcpu/mod.rs
```diff
@@ -57,6 +57,7 @@ pub type Result<T> = result::Result<T, Error>;
 /// This struct is a temporary (and quite terrible) placeholder until the
 /// [`vmm-vcpu`](https://github.com/rust-vmm/vmm-vcpu) crate is stabilized.
 pub(crate) struct Vcpu {
+    id: u8,
```
aghecenco: Should we make this `usize` and save ourselves future extensions for >255 vCPUs?

src/vmm/src/vm.rs
```diff
@@ -0,0 +1,105 @@
+use crate::vcpu::{self, Vcpu};
```
aghecenco: Missing copyright.

src/vmm/src/vm.rs
```diff
@@ -0,0 +1,105 @@
...
+use kvm_bindings::{kvm_pit_config, kvm_userspace_memory_region, KVM_PIT_SPEAKER_DUMMY};
+use kvm_ioctls::{Kvm, VmFd};
+use std::sync::Arc;
```
aghecenco: nit: reorder these `use`s to separate `std` from internal crates and everything from extern crates

src/vmm/src/vm.rs
```diff
@@ -0,0 +1,105 @@
...
+impl KvmVm {
+    /// Create a new `KvmVm`.
+    pub fn new(kvm: &Kvm) -> Result<Self> {
+        let fd = kvm.create_vm().map_err(Error::CreateVm)?;
+        Ok(KvmVm {
+            fd,
+            vcpus: Vec::new(),
+        })
+    }
+    /// Set the user memory region.
```
aghecenco: nit: add newline

src/vmm/src/vm.rs
```diff
@@ -0,0 +1,105 @@
...
+        &self,
+        memory_region: kvm_userspace_memory_region,
+    ) -> Result<()> {
+        self.fd
+            .set_user_memory_region(memory_region)
+            .map_err(Error::SetupMemoryRegion)
+    }
+
+    /// Configures the in kernel interrupt controller.
+    // This function should be reuse to configure the aarch64 interrupt controller (GIC).
```
aghecenco:
```suggestion
    // This function should be reused to configure the aarch64 interrupt controller (GIC).
```

src/vmm/src/vm.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    /// Let KVM know that instead of triggering an actual interrupt for `irq_number`, we will
+    /// just write on the specified `event`.
+    pub fn register_irqfd(&self, event: &EventFd, irq_number: u32) -> Result<()> {
+        self.fd
+            .register_irqfd(&event, irq_number)
+            .map_err(Error::RegisterIrqEvent)
+    }
+
+    pub fn run(&mut self) {
```
aghecenco: nit: missing documentation

src/vmm/src/vm.rs
```diff
@@ -27,6 +33,7 @@ pub enum Error {
     CreateVcpu(vcpu::Error),
     /// Failed to register IRQ event.
     RegisterIrqEvent(kvm_ioctls::Error),
+    RunVcpus(io::Error),
```
aghecenco: Missing documentation.

src/vmm/src/lib.rs
```diff
@@ -41,7 +41,9 @@ use devices::SerialWrapper;
 
 mod vcpu;
 mod vm;
-use vcpu::{cpuid, mpspec, mptable, Vcpu};
+use crate::vcpu::cpuid::filter_cpuid;
+use crate::vcpu::VcpuState;
```
aghecenco: Is `crate::` necessary?

src/vmm/src/lib.rs
```diff
@@ -296,44 +298,41 @@ impl VMM {
         Ok(())
     }
 
-    /// Configure guest vCPUs.
+    /// Creates guest vCPUs.
```
aghecenco, line 263: nit (active voice)
```suggestion
    /// Create guest vCPUs.
```

src/vmm/src/vcpu/mod.rs
```diff
@@ -52,39 +52,54 @@ pub enum Error {
 /// Dedicated Result type.
 pub type Result<T> = result::Result<T, Error>;
 
+pub(crate) struct VcpuState {
```
aghecenco: Not sure about calling it `*State`, as state conceptually means what a vCPU is doing at a point in time, like what's in its registers etc. `VcpuConfig` is taken for the user-facing part of the VMM but this is still more of a initialization/configuration ephemeral object.

src/vm-vcpu/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+authors = ["rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+libc = "0.2.76"
+kvm-bindings = { version = "0.3.0", features = ["fam-wrappers"] }
+kvm-ioctls = "0.5.0"
+vm-device = { git = "https://github.com/aghecenco/vm-device.git", branch = "send" }
```
aghecenco, line 13: This is prettier with the `[patch]` approach seen here: https://github.com/rust-vmm/vmm-reference/pull/32/commits/d0318fc093e71debef1d6bf8fee0911d23b57c1e

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -52,17 +53,20 @@ pub enum Error {
 /// Dedicated Result type.
 pub type Result<T> = result::Result<T, Error>;
 
-pub(crate) struct VcpuState {
+pub struct VcpuState {
     pub id: u8,
     pub cpuid: CpuId,
     pub kernel_load_addr: GuestAddress,
+    // todo: make the load_addr and zero_page_start addr uniform.
+    // they should either both be u64 or GuestAddress.
```
aghecenco, line 59: I vote for `GuestAddress`.

src/vm-vcpu/Cargo.toml
```diff
@@ -11,5 +11,8 @@ libc = "0.2.76"
 kvm-bindings = { version = "0.3.0", features = ["fam-wrappers"] }
 kvm-ioctls = "0.5.0"
 vm-device = { git = "https://github.com/aghecenco/vm-device.git", branch = "send" }
-vm-memory = { version = "0.3.0", features = ["backend-mmap"] }
+vm-memory = { version = "0.3.0" }
 vmm-sys-util = "0.6.1"
+
+[dev-dependencies]
+vm-memory = { version = "0.3.0", features = ["backend-mmap"]}
```
aghecenco, line 18: nit: add missing newline

src/vmm/src/lib.rs
```diff
@@ -118,43 +116,42 @@ pub struct VMM {
...
             vm,
             kvm,
-            guest_memory: GuestMemoryMmap::default(),
+            guest_memory,
             device_mgr: Some(IoManager::new()),
             event_mgr: EventManager::new().map_err(Error::EventManager)?,
         };
 
-        vmm.check_kvm_capabilities()?;
+        vmm.configure_pio_devices()?;
```
aghecenco, line 145: Why is this here?

src/vm-vcpu/src/vm.rs
```diff
@@ -0,0 +1,285 @@
...
+/// A KVM specific implementation of a Virtual Machine.
+///
+/// Provides abstractions for working with a VM. Once a generic Vm trait will be available,
+/// this type will become on of the concrete implementations.
+pub struct KvmVm {
+    fd: VmFd,
+    state: VmState,
+    // Only one of `vcpus` or `vcpu_handles` can be active at a time.
+    // To create the `vcpu_handles` the `vcpu` vector is drained.
+    // A better abstraction should be used to represent this behavior.
```
acatangiu, line 27: [This model](https://github.com/firecracker-microvm/firecracker/blob/0380c2c6b0dd3783e16630eacd2a12c79de72439/src/vmm/src/vstate/vcpu/mod.rs#L227) simplified things a lot in `firecracker`.

src/vmm/src/vcpu/mod.rs
andreeaflorescu: The reason for calling it `State` is that I was looking into how we are going to implement features such as pause & resume. The general pattern that I was thinking about applying here is that the `Vcpu` and `Vm` should have only one method of initialization (i.e. new). Then, the only thing that differs from when you are starting a fresh Vcpu versus one that was resume is the `State`.

Right now since we didn't really look at the state that needs to be saved on `pause`, the vcpu & vm state is really basic, but as we add this implementation I am expecting the `State` structure to evolve such that:
- you can easily create a `default`/`empty` state; this is to be used when starting a new vm/vcpu
- add other fields to the structure related to other registers/other things that need to be saved; this should be used when resuming from a known state.

Does that make sense?

src/vm-vcpu/Cargo.toml
andreeaflorescu: I think we can now use the upstream one. I'll update it.

src/vm-vcpu/src/vm.rs
```diff
@@ -0,0 +1,285 @@
...
+    vcpus: Vec<KvmVcpu>,
```
aghecenco, line 28: I'm not sure about storing the vCPUs in the `KvmVm` object. I don't see how they're related. One abstracts the hypervisor (`Vm`, currently `KvmVm` as it doesn't abstract it all the way) and the other abstracts something else entirely, the guest's CPUs. Moreover, the hypervisor is something you want to pass around later on in the execution of your program - for instance, devices need it at boot time to get activated (currently, that translates into registering the irqfd with KVM). I don't like sending the vCPUs too just for the sake of having fewer members in the `vmm` struct.

alexandruag: I guess one way to look at this is `KvmVm` being an standalone abstraction that performs global and vcpu setup (and later start) based on KVM, behind an interface that provides as much convenience as possible. To that end, I think having the vcpus here makes sense; otherwise, as this becomes a separate component, each consumer would have to explicitly store them somewhere as well. It def seems useful to have accessors for some of the members, and I guess we'll get a better idea as we starting using these things and build on top of them.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -66,23 +66,23 @@ pub struct VcpuState {
...
     vcpu_fd: VcpuFd,
     /// Device manager for bus accesses.
     device_mgr: Arc<IoManager>,
     state: VcpuState,
 }
 
-impl Vcpu {
-    /// Create a new vCPU.
+impl KvmVcpu {
+    /// Create a new vCPU.127:46
```
alexandruag, line 76: "127:46"

src/vm-vcpu/src/vm.rs
```diff
@@ -44,29 +55,59 @@ pub type Result<T> = std::result::Result<T, Error>;
...
             vcpus: Vec::new(),
             vcpu_handles: Vec::new(),
-        })
+        };
+
+        vm.configure_memory_regions(guest_memory, kvm)?;
+
+        mptable::setup_mptable(guest_memory, vm.state.num_vcpus).map_err(Error::Mptable)?;
+
+        vm.setup_irq_controller()?;
```
alexandruag, line 72: Does setting up the irq controller here creates issues on Arm (I think there was that discussion that vCPUs have to be created first on that platform). If so, maybe one way around this going forward is to create just the raw vcpu fds to begin with for Arm, and set up the proper vcpu wrappers later.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: It's not my pin, I promise.

src/vm-vcpu/src/vm.rs
andreeaflorescu: I was considering taking a different approach with aarch64 actually. Most of the calls in this function do not make sense for aarch64, but I was thinking of reusing function names, and just call them at the appropriate time depending on the platform. I think I can better explain this in the code directly. For now I would leave the setup here, and on aarch64 the irq setup should transparently be called when creating the vcpus.

src/vmm/src/lib.rs
```diff
@@ -324,32 +310,30 @@ impl VMM {
...
-
             // Set CPUID.
+            self.vm.create_vcpu(index, shared_device_mgr.clone())?;
+        }
+        // Set CPUID.
+        let base_cpuid = self
+            .kvm
+            .get_supported_cpuid(KVM_MAX_CPUID_ENTRIES)
+            .map_err(Error::KvmIoctl)?;
+        for vcpu in self.vm.vcpus.iter() {
```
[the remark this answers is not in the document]
andreeaflorescu: While it will greatly simplify things in this particular use case, we would still need to offer functionality for setting a custom cpuid because the filtering is application specific. If we keep things as they are now then Firecracker would be able to consume the vm-vcpu crate and provide a cpuid that corresponds to the Firecracker specific templates (T2/C3). I would rather prefer to have something like creating a default cpuid to simplify the code here instead of moving this call in `Vm`. What do you say?

src/vmm/src/lib.rs
```diff
@@ -126,10 +132,10 @@ impl VMM {
         }
 
         // Create fd for interacting with kvm-vm specific functions.
-        let vm_fd = kvm.create_vm().map_err(Error::KvmIoctl)?;
+        let vm = KvmVm::new(&kvm)?;
```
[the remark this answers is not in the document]
andreeaflorescu: Conceptually this doesn't really make sense to me. Is it simplifying anything if we're doing this?

src/vmm/src/vm.rs
```diff
@@ -0,0 +1,105 @@
...
+use std::thread;
+use vm_device::device_manager::IoManager;
+use vmm_sys_util::eventfd::EventFd;
+
+/// A KVM specific implementation of a Virtual Machine.
+///
+/// Provides abstractions for working with a VM. Once a generic Vm trait will be available,
+/// this type will become on of the concrete implementations.
+pub struct KvmVm {
+    fd: VmFd,
```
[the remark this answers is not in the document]
andreeaflorescu: I'll have to think a bit about this one. I don't particularly like the approach because it is leaking implementation details about the VM in the devices, and I am thinking about maybe finding a better abstraction there.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: Not only are they linux specific, but they're also a bit linux-loader specific (because we do not have a firmware that sets the registers up). We're probably better off having something like passing the registers directly (which is also why I was a bit on the fence weather to use GuestAddress or u64, since the registers are using u64 for their definition).

src/vm-vcpu/src/vm.rs
andreeaflorescu: I think at the conceptual level the `vcpus` belong to the `Vm` abstraction because the VM should represent a machine with cpus & memory. Since there was no hurdle to actually move the ownership of the Vcpus to the Vm, I just went ahead and did that. This allows us to have `vm.run` that creates threads for the vcpus, and runs vcpu run. It is also the entity responsible for the lifetime of vcpus, and can do things such as joining the threads, and in the future kicking the vcpus when needed.

andreeaflorescu (commented):
I'll address the comments, and ping you afterwards.

src/vmm/src/vcpu/mod.rs
andreeaflorescu: Right now we cannot make this an u32 because the kvm-ioctls crate expects an u8 here. We first need to update the kvm-ioctls crate, and then we can also increase this number. Btw, KVM API defines this as an u64, so I think we should just do the same. I've opened an issue for this (https://github.com/rust-vmm/kvm-ioctls/issues/123), but I wouldn't block on the kvm-ioctls version to be published if you're also okay with that.

src/vmm/src/vm.rs
andreeaflorescu: I kinda changed my mind and now the comment below is not accurate anymore, so I will just remove it :smile:

src/vmm/src/lib.rs
andreeaflorescu: I have no idea why, but in the vm-vcpu crate, if I do not write `crate::` it cannot find the vcpu module :(. I left it as is.

src/vmm/src/vm.rs
andreeaflorescu: Done.

src/vmm/src/vm.rs
andreeaflorescu: Done.

src/vm-vcpu/Cargo.toml
andreeaflorescu: I added a new commit which updates the vm-device dependency everywhere to use the upstream version.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: Updated zero_page_start to GuestAddress.

src/vmm/src/lib.rs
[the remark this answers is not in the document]
alexandruag: Yeah I guess it's ultimately debatable who should own the `Kvm` object. However, things like checking for capabilities seem like they should be `KvmVm`-specific.

src/vmm/src/lib.rs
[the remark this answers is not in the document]
alexandruag: The input (i.e. cpuid config) we use here to create each vcpu could be passed to `KvmVm` as well so it can handle things behind the scenes. For now we can proceed with either approach and see how things shape up.

src/vmm/src/vcpu/mod.rs
alexandruag: That's fine with me.

src/vmm/src/vm.rs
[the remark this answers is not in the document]
alexandruag: There are kinda two separate issues here. Holding the fd as an `Arc<VmFd>` and providing access from the outside is an overall useful interface addition IMO. On the device side, right now the top level devices also make some specific assumptions (i.e. we're using `kvm` and things like `EventFd`s), and then they also incorporate some building blocks which are already shaping up to be generic. Also, I guess this second thing is something we can talk about on the device PR.

src/vm-vcpu/src/vm.rs
alexandruag: Sure we can clarify this later.

alexandruag (approved):

aghecenco (approved):

aghecenco (approved):

alexandruag (approved):

PR #38, opened 2020-11-23
Refactoring - use default & de-duplicate names

andreeaflorescu:
Use default values for command line arguments so that we can enable a simple usage such as:
```vmm-reference --kernel path=/path/to/kernel/image```

Updated the rust-vmm-ci as well.

### CODE REVIEW

File: Cargo.toml [hunk 1]
```diff
@@ -7,6 +7,6 @@ license = "Apache-2.0 OR BSD-3-Clause"
 
 [dependencies]
 vmm = { path = "src/vmm" }
-cli = { path = "src/cli" }
+cli = { path = "src/api" }
```
aghecenco, line 10:
```suggestion
api = { path = "src/api" }
```

aghecenco: This implies that you also rename `package` to `api` in `api/Cargo.toml` and that you should change `cli` to `api` in `main.rs`. `Cargo.lock` will also need to be updated.

aghecenco (commented):
nit: typo in commit message for bd8f24897c30a752381ea7c8b80612f73dd54c48: s/argumetns/arguments

aghecenco:
I'm not sure the arguments are parsed correctly, I think the space separator that replaces the comma is confusing the `ArgParser`.

This command line:
```
target/debug/vmm-reference \
    --memory size_mib=1024 \
    --kernel cmdline="console=ttyS0 i8042.nokbd reboot=k panic=1 pci=off" path=resources/kernel/bzimage-hello-busybox \
    --vcpu num=1
```
doesn't appear to pass `path` to the `KernelConfig` object:

```
Failed to parse command line options. Invalid command line arguments: error: Found argument 'path=resources/kernel/bzimage-hello-busybox' which wasn't expected, or isn't valid in this context

USAGE:
    vmm-reference --kernel <kernel> --memory <memory> --vcpu <vcpu>

For more information try --help
```

src/vmm/src/config.rs
```diff
@@ -103,12 +109,9 @@ impl TryFrom<String> for KernelConfig {
...
         // Supported options:
-        // `cmdline=<"string">,path=/path/to/kernel,himem_start=<u64>`
-        // Required: all
-        let options: Vec<&str> = kernel_cfg_str.split(',').collect();
-        if options.len() != 3 {
-            return Err(ConversionError::ParseKernel(kernel_cfg_str));
-        }
+        // `cmdline=<"string"> path=/path/to/kernel himem_start=<u64>`
+        // Required: path
+        let options: Vec<&str> = kernel_cfg_str.split(' ').collect();
```
lauralt, line 114: This will split the `cmdline` value in its constituents too, which will make the following split/parsing pretty difficult. Can we stick to `,`?

src/vmm/src/lib.rs
```diff
@@ -57,6 +57,12 @@ const ZEROPG_START: u64 = 0x7000;
 /// Address where the kernel command line is written.
 const CMDLINE_START: u64 = 0x0002_0000;
 
+/// Default high memroy start.
```
lauralt, line 60: typo: `memroy`.

src/vmm/src/lib.rs
```diff
@@ -57,6 +57,12 @@ const ZEROPG_START: u64 = 0x7000;
...
+pub const HIGH_RAM_START: u64 = 0x100000;
+
+/// Default kernel command line.
```
lauralt, line 63: nit: probably not in the scope of this PR but it would be nice to document a bit why we chose this default cmdline (or just open an issue for that).

src/cli/src/lib.rs
```diff
@@ -32,21 +32,23 @@ impl CLI {
                     .long("memory")
                     .required(true)
                     .takes_value(true)
+                    .default_value("size_mib=256")
```
lauralt, line 35: The docs will need to be updated with some info about the default values and, if we change the separators (or other things), those will have to be updated there too.

andreeaflorescu: The default value is already printed by the default help.

This is how the help looks like with these changes:
```bash

USAGE:
    target/debug/vmm-reference --kernel <kernel> --memory <memory> --vcpu <vcpu>

FLAGS:
    -h, --help       
            Prints help information

    -V, --version    
            Prints version information


OPTIONS:
        --kernel <kernel>    
            Kernel configuration.
            	Format: "path=<string> [cmdline=<string> himem_start=<u64>]"
        --memory <memory>    
            Guest memory configuration.
            	Format: "size_mib=<u32>" [default: size_mib=256]
        --vcpu <vcpu>        
            vCPU configuration.
            	Format: "num=<u8>" [default: num=1]

```

For the kernel configuration I didn't find a good way of printing the help and also denote that some of the config options are optional.

andreeaflorescu: Do you mean other documents rather than the help?

src/vmm/src/lib.rs
andreeaflorescu: I am sort of starting to believe that the command line should not be exposed to the user in any way because it is very hard to configure it correctly. Instead of being able to configure the command line, we should allow users to configure options in a higher level language. For example, the typical usage of the command line is to disable the serial console device. Instead of allowing modification on the kernel command line, we can just add a flag parameter (`--serial`) and update the cmdline under the hood without direct user interaction. Does that make sense?

src/vmm/src/config.rs
andreeaflorescu: Yap, that makes total sense. I completely missed that. I'll revert it to use `,`.

src/cli/src/lib.rs
lauralt: Yep, sorry, I was referring to DESIGN.md and getting-started, the command examples could be updated, for example the one from [here](https://github.com/rust-vmm/vmm-reference/blob/master/docs/DESIGN.md#user-interface), maybe the `Inputs` from [here](https://github.com/rust-vmm/vmm-reference/blob/master/docs/DESIGN.md#functionalities) too (with mandatory/not mandatory or their defaults for example).

andreeaflorescu: Okappa, will update.

andreeaflorescu: I've updated the readme and added 2 more examples. For the usage example i chose the most basic example.

In the design document, I chose to remove the "user defined" mention because otherwise we would end up with duplicated information in two places which I believe makes things a bit harder to maintain.

lauralt: :+1: These [lines](https://github.com/rust-vmm/vmm-reference/blame/master/docs/DESIGN.md#L158-L161) and the examples from [here](https://github.com/rust-vmm/vmm-reference/blob/master/docs/getting-started.md#putting-it-all-together) will also need an update.

Also, totally unrelated to this PR, just noticed that we didn't update the documentation related to all the changes introduced in the bzimage PR :(.

src/vmm/src/config.rs
```diff
@@ -104,11 +108,8 @@ impl TryFrom<String> for KernelConfig {
     fn try_from(kernel_cfg_str: String) -> result::Result<Self, Self::Error> {
         // Supported options:
         // `cmdline=<"string">,path=/path/to/kernel,himem_start=<u64>`
-        // Required: all
+        // Required: path
         let options: Vec<&str> = kernel_cfg_str.split(',').collect();
```
lauralt, line 112:
```suggestion
        let options: Vec<&str> = kernel_cfg_str.split(',').filter(|tok| !tok.is_empty()).collect();
```

Otherwise, we panic if some option is an empty string, for example one comma at the end:

```bash
cargo run --                                                               
              --memory size_mib=1024 
              --vcpu num=1                     
              --kernel path=vmlinux-hello-busybox,himem_start=1024,cmdline="console=ttyS0 i8042.nokbd reboot=k panic=1 pci=off",
```

src/main.rs
```diff
@@ -1,7 +1,7 @@
 // Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
-extern crate cli;
+extern crate api;
 extern crate vmm;
```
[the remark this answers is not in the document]
andreeaflorescu: I've added a new commit to remove all `extern crate` defines.

src/vmm/src/lib.rs
lauralt: Sounds nice.

src/cli/src/lib.rs
andreeaflorescu: I've updates those as well, but realized that the documentation is behind by a large margin. We need to go back and update it. For the programmatic interface, I would wait a bit more, and I am also considering not mentioning function names & such because those are always in flux. For code level documentation, people can do a cargo doc --open on the vmm crate, so we could encourage them to do that instead (or even host something like that). Then we can move the documentation in the crate itself so that the examples are also built which increases the changes of the documentation to stay up to date. @aghecenco what do you think?

In the meantime, I've open this issue: #39

lauralt (approved):

src/vmm/src/config.rs
```diff
@@ -157,3 +161,31 @@ pub struct VMMConfig {
...
     pub kernel_config: KernelConfig,
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_kernel_config() {
+        // Check that additional comas in the kernel string do not cause a panic.
```
aghecenco, line 171: nit: s/comas/commas

aghecenco (approved):

src/vmm/src/config.rs
andreeaflorescu: :scream:

PR #40, opened 2020-11-27
More unit tests for vmm configuration

### LINKED ISSUE #15: Add more unit tests & negative tests for glue code
The "glue" part does a bit more than stitch stuff together: VMM-specific configurations like boot parameters or register values are also set up here. The code that handles this should be adequately unit tested. In time, some of it can migrate to other crates, and it should do so with tests and all.

* [ ] guest memory: #28
* [ ] kernel: #29
* [ ] vCPUs: #30
* [ ] boot parameters: #31

aghecenco:
Fixes #15

### DIFF
--- a/src/vmm/src/boot.rs
+++ b/src/vmm/src/boot.rs
@@ -28,15 +28,13 @@ const EBDA_START: u64 = 0x0009_fc00;
 // See https://github.com/rust-vmm/linux-loader/issues/51
 const E820_RAM: u32 = 1;
 
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 /// Errors pertaining to boot parameter setup.
 pub enum Error {
     /// Invalid E820 configuration.
     E820Configuration,
     /// Highmem start address is past the guest memory end.
     HimemStartPastMemEnd,
-    /// Highmem start address is past the MMIO gap end.
-    HimemStartPastMmioGapEnd,
     /// Highmem start address is past the MMIO gap start.
     HimemStartPastMmioGapStart,
     /// The MMIO gap end is past the guest memory end.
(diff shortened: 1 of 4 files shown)


src/vmm/src/lib.rs
```diff
@@ -374,17 +374,20 @@ impl VMM {
 mod tests {
     use super::*;
 
+    use std::fs;
```
andreeaflorescu: This does not seem to be used.

src/vmm/src/lib.rs
```diff
@@ -374,17 +374,20 @@ impl VMM {
...
     use std::io::ErrorKind;
+    use std::mem;
+    use std::os::unix::fs::PermissionsExt;
```
andreeaflorescu: This does not seem to be used either.

src/vmm/src/lib.rs
```diff
@@ -512,10 +525,21 @@ mod tests {
...
         vmm_config.kernel_config.path = PathBuf::from(temp_file.as_path());
         let mut vmm = mock_vmm(vmm_config);
+        assert!(matches!(
+            vmm.load_kernel().unwrap_err(),
+            Error::KernelLoad(loader::Error::Bzimage(
+                loader::bzimage::Error::InvalidBzImage
+            ))
+        ));
 
+        // Test case: kernel path doesn't point to a file.
```
andreeaflorescu, line 532: It looks like an empty file would yield the same error. Do we need a directory?

src/vmm/src/lib.rs
```diff
@@ -604,4 +603,29 @@ mod tests {
             if e.kind() == ErrorKind::InvalidInput
         ));
     }
+
+    #[test]
+    fn test_create_vcpus() {
+        // The scopes force the created vCPUs to unmap their kernel memory at the end.
```
andreeaflorescu, line 607: Unmapping the kernel memory is not needed since we are recreating the mock vmm.

src/vmm/src/boot.rs
```diff
@@ -133,17 +131,146 @@ pub fn build_bootparams(
...
     Ok(params)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    use linux_loader::bootparam;
+
+    const HIGH_RAM_START: u64 = 0x0010_0000;
```
andreeaflorescu, line 156: This can be imported instead of being redefined:
```use crate::HIGH_RAM_START;```.

File: src/vmm/src/boot.rs [hunk 6]
```diff
@@ -133,17 +131,146 @@ pub fn build_bootparams(
...
+    // First address past 32 bits is where the MMIO gap ends.
+    const MMIO_GAP_END: u64 = 1 << 32;
+    // Size of the MMIO gap.
+    const MMIO_GAP_SIZE: u64 = 768 << 20;
+    // The start of the MMIO gap (memory area reserved for MMIO devices).
+    const MMIO_GAP_START: u64 = MMIO_GAP_END - MMIO_GAP_SIZE;
```
andreeaflorescu, line 162: How about we define these as ```pub (crate)``` in vmm/src/lib.rs and then just import them instead of redefining them?

src/vmm/src/lib.rs
```diff
@@ -501,7 +511,10 @@ mod tests {
 
         // Test case: kernel file does not exist.
         let mut vmm_config = default_vmm_config();
-        vmm_config.kernel_config.path = PathBuf::from("not_a_file");
+        {
+            let temp_file = TempFile::new().unwrap();
+            vmm_config.kernel_config.path = PathBuf::from(temp_file.as_path());
+        }
```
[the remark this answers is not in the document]
aghecenco: It's paranoia against future us who might leave a `not_a_file` hanging around there. I want to be as sure as possible that the file doesn't exist, hence the `tempfile` that gets deleted when it goes out of scope.

src/vmm/src/lib.rs
aghecenco: Nope, it's the same code path so we gain nothing on coverage, but it's an edge case nonetheless.

src/vmm/src/boot.rs
aghecenco: They're defined a bit different in `vmm` -
```rust
/// First address past 32 bits.
const FIRST_ADDR_PAST_32BITS: u64 = 1 << 32;
/// Size of the MMIO gap.
const MEM_32BIT_GAP_SIZE: u64 = 768 << 20;
/// The start of the memory area reserved for MMIO devices.
const MMIO_MEM_START: u64 = FIRST_ADDR_PAST_32BITS - MEM_32BIT_GAP_SIZE;
```
and I thought this way was more straightforward. Would it be OK to redefine the ones in `vmm/lib.rs` as `MMIO_GAP_START`, `_SIZE` and `_END` instead of the above, and export them?

src/vmm/src/lib.rs
aghecenco: When we recreate it, the object isn't destroyed, it's shadowed (Rust lets us create a new object with the same name and doesn't let us access the old one anymore within the same scope, but *doesn't* immediately clean it up). I put a breakpoint in `KvmRunWrapper::drop` and it's never called until the end of the function, unless we explicitly enforce scopes. We can of course leave the memory there (I'm not sure if it `ENOMEM`s at the 256th vcpu) but I prefer the forced cleanup

src/vmm/src/boot.rs
andreeaflorescu: I would use the values from vmm, and rename them if needed. Defining the constants twice is a bit confusing IMHO.

aghecenco: Done

andreeaflorescu (approved):

src/vmm/src/lib.rs
```diff
@@ -46,19 +46,19 @@ pub use config::*;
...
+pub(crate) const MMIO_GAP_START: u64 = MMIO_GAP_END - MMIO_GAP_SIZE;
 /// Address of the zeropage, where Linux kernel boot parameters are written.
 const ZEROPG_START: u64 = 0x7000;
 /// Address where the kernel command line is written.
 const CMDLINE_START: u64 = 0x0002_0000;
 
-/// Default high memory start.
-pub const HIGH_RAM_START: u64 = 0x100000;
+/// Default high memory start (1 MiB).
+pub const HIGH_RAM_START: u64 = 0x0010_0000;
```
lauralt, line 61: nit: not necessarily related to the changes in this PR, but I would add the `DEFAULT` prefix to HIGH_RAM_START too, otherwise looks like the high memory always starts at this address.

src/vmm/src/lib.rs
```diff
@@ -432,15 +432,25 @@ mod tests {
         }
     }
 
+    // Return the address where an ELF file should be loaded, as specified in its header.
+    fn elf_load_addr(elf_path: &PathBuf) -> GuestAddress {
+        let mut elf_file = File::open(elf_path.as_path()).unwrap();
```
lauralt, line 437: nit: the `as_path()` is not needed.

src/vmm/src/boot.rs
```diff
@@ -133,17 +131,139 @@ pub fn build_bootparams(
             add_e820_entry(
                 &mut params,
                 mmio_gap_end.raw_value(),
+                // The unchecked_offset_from is safe because:
+                // * mmio_gap_end > mmio_gap_start, validated at the beginning of the function
+                // * mmio_gap_start > himem_start, validated above
+                // therefore, mmio_gap_end > himem_start
```
lauralt, line 137: Isn't it enough to say the `unchecked_offset_from` is safe because last_addr > mmio_gap_end (true because of the above `if`)?

src/vmm/src/boot.rs
```diff
@@ -133,17 +131,139 @@ pub fn build_bootparams(
...
+
+    use crate::{HIGH_RAM_START, MMIO_GAP_END, MMIO_GAP_START};
+    use linux_loader::bootparam;
+
+    #[test]
+    fn test_build_bootparams() {
+        let guest_memory = GuestMemoryMmap::default();
+        let mut kern_load_res = KernelLoaderResult::default();
+
+        // Error case: MMIO gap end address is past its start address.
```
lauralt, line 162: Shouldn't this be rephrased to `Error case: MMIO gap start address is past its end address.`?

andreeaflorescu (approved):

lauralt (approved):

PR #41, opened 2020-11-29
Add block device support

alexandruag:
This PR comes as a follow up on #32, for easier review/comparison due to incorporating multiple changes. We now use the request parsing and execution functionality implemented in https://github.com/rust-vmm/vm-virtio/pull/29. I've added some unit tests, but it seems most of the logic here should be exercised as part of integration tests (which have to be added as well, but it would be great if that was a separate thread we can conclude as the net functionality is added on top of this). The last commit implements support for both net and block configuration via the command line, but only the latter is actually used for now.


src/devices/src/virtio/block/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+    use vmm_sys_util::tempfile::TempFile;
+
+    use super::*;
+    use std::convert::TryInto;
+
+    #[test]
+    fn test_build_config_space() {
+        let tmp = TempFile::new().unwrap();
+
+        // TODO: There no way to get (via `Deref` or something else) the file associated with
```
lauralt, line 87: into_file() should do the job.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -266,9 +266,12 @@ impl KvmVcpu {
                             {
                                 eprintln!("Failed to write to serial port");
                             }
-                        } else if addr == 0x060 || addr == 0x061 || addr == 0x064 {
-                            // Write at the i8042 port.
-                            // See https://wiki.osdev.org/%228042%22_PS/2_Controller#PS.2F2_Controller_IO_Ports
+                        } else if data.len() == 1 && data[0] == 0xfe && addr == 0x64 {
```
lauralt, line 269: I think we should open an issue to not forget about using i8042 from vm-superio once we do a new release there (assuming we are waiting for the metrics to be merged until creating a new release).

src/devices/src/virtio/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+#[derive(Copy, Clone)]
+pub struct MmioConfig {
+    pub range: MmioRange,
+    // The interrupt assigned to the device.
+    pub gsi: u32,
+}
+
+// These arguments are common for all virtio devices. We're always passing a mmio_cfg object
+// for now, and we'll re-evaluate the layout of this struct when adding more transport options.
+pub struct CommonArgs<'a, M, B> {
```
lauralt, line 49: I wonder if we can find another word instead of `args` (applies to `BlockArgs` as well). `args` makes me think of cmd line arguments :-?

src/vmm/src/lib.rs
```diff
@@ -301,6 +311,34 @@ impl VMM {
         Ok(())
     }
 
+    fn add_block_device(&mut self, cfg: &BlockConfig) -> Result<()> {
```
lauralt, line 314: Is it too early to add other parameters to this (for example the `MmioConfig`) or to split it in 2 fns and make it, this way, work for any device?

lauralt (commented):
Just a few things I noticed while scrolling. I'll take a closer look at it soon.

src/devices/src/virtio/block/queue_handler.rs
```diff
@@ -0,0 +1,54 @@
...
+// to interact with the event manager. `ioeventfd` is the `EventFd` connected to queue
+// notifications coming from the driver.
+pub(crate) struct QueueHandler<M: GuestAddressSpace> {
+    pub inner: InOrderQueueHandler<M, SingleFdSignalQueue>,
+    pub ioeventfd: EventFd,
+}
+
+impl<M: GuestAddressSpace> MutEventSubscriber for QueueHandler<M> {
+    fn process(&mut self, events: Events, _ops: &mut EventOps) {
+        if events.event_set() != EventSet::IN {
```
andreeaflorescu, line 26: Should we use `contains` instead of `!=`?

You can receive error & hup events as well. Should we deregister the event in that case? Or register some sort of event for error handling that can be then processed by the vmm itself?

Nevertheless, we should not continue the processing when the event_set is not IN.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: Do we even need this patch for anything?

Cargo.toml
```diff
@@ -18,3 +18,9 @@ panic = "abort"
 codegen-units = 1
 lto = true
 panic = "abort"
+
+[patch.crates-io]
+event-manager = { git = "https://github.com/alexandruag/event-manager.git", branch = "fix" }
```
andreeaflorescu, line 23: Can you please post the PR in event-manager as well? We already have a few forks that we need to keep track of.

src/vmm/src/lib.rs
```diff
@@ -301,6 +311,34 @@ impl VMM {
...
+            kernel_cmdline: &mut self.kernel_cfg.cmdline,
+        };
+
+        let args = BlockArgs {
+            common,
+            file_path: PathBuf::from(&cfg.path),
+        };
+
+        // We can also hold this somewhere if we need to keep the handle for later.
+        let _b = Block::new(args).expect("failed to create block");
```
andreeaflorescu, line 337: We should return an error here and not use expect.

src/vmm/src/lib.rs
```diff
@@ -157,8 +163,12 @@ impl TryFrom<VMMConfig> for VMM {
             kernel_cfg: config.kernel_config.clone(),
         };
 
-        vmm.add_serial_console()?;
         vmm.create_vcpus(&config.vcpu_config)?;
+        vmm.add_serial_console()?;
+
+        if config.block_config.is_some() {
+            vmm.add_block_device(&config.block_config.unwrap()).unwrap();
```
andreeaflorescu, line 170: We shouldn't unwrap here and propagate the error. This allows us to write negative tests as well as printing nice errors to the user.

src/devices/src/virtio/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+                Arc::new(GuestMemoryMmap::from_ranges(&[(GuestAddress(0), 0x1000_0000)]).unwrap());
+
+            let kvm = kvm_ioctls::Kvm::new().unwrap();
+            let vm_fd = Arc::new(kvm.create_vm().unwrap());
+
+            let range = MmioRange::new(MmioAddress(0x1_0000_0000), 0x1000).unwrap();
+            let mmio_cfg = MmioConfig { range, gsi: 5 };
+
+            // Required so the vm_fd can be used to register irqfds.
+            vm_fd.create_irq_chip().unwrap();
```
andreeaflorescu, line 127: This code will need to be duplicated in the vmm crate as well if we want to test `add_block_device`, and there we should remove this call to `create_irq_chip` because Vm::new() is expected to be called. I am wondering if it makes sense to pass the VmFd as a parameter instead, and later on have a test_utilities feature behind which we can hide this functionality.

src/devices/src/virtio/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+    use vm_device::bus::MmioAddress;
+    use vm_device::device_manager::IoManager;
+    use vm_memory::{GuestAddress, GuestMemoryMmap};
+
+    use super::*;
+
+    // Can be used in other modules to test functionality that requires a `CommonArgs` struct as
+    // input. The `args` method below generates an instance of `CommonArgs` based on the members
+    // below.
+    pub struct CommonArgsContext {
```
andreeaflorescu, line 106: Can we call it `CommonArgsMock` instead as it is not configurable in anyway?

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,263 @@
...
+
+#[cfg(test)]
+mod tests {
+    use vmm_sys_util::tempfile::TempFile;
+
+    use crate::virtio::tests::CommonArgsContext;
+
+    use super::*;
+
+    // Restricting this for now, because registering irqfds does not work on Arm without properly
```
andreeaflorescu, line 229: Besides the bus range check anything else looks very x86_64 specific at the moment. To make it less specific we would need to abstract the way we specify devices (cmdline on x86), and update the CommonArgsContext such that it does not also create the irqchip. It should instead receive just a reference to the VmFd and assume that the irqchip is already created so that we can reuse the code on arm as well.

Since it is not really reusable at the moment, I would just make this test x86_64 specific with `cfg(target_arch = x86_64)`.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -288,7 +291,12 @@ impl KvmVcpu {
                                 eprintln!("Failed to read from serial port");
                             }
                         } else {
-                            // Read from some other port.
+                            // Emulate a dummy i8042.
+                            if data.len() == 1 && addr == 0x64 {
+                                data[0] = 0x0;
+                            } else if data.len() == 1 && addr == 0x61 {
+                                data[0] = 0x20;
```
aghecenco, line 298: Can you please add a comment explaining why `0x20`?

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,263 @@
...
+            file_path: args.file_path,
+        }));
+
+        // Register the device on the MMIO bus.
+        args.common
+            .mmio_mgr
+            .register_mmio(mmio_cfg.range, block.clone())
+            .map_err(Error::Bus)?;
+
+        // TODO: We're hardcoding the MMIO range len to 4K so assert for this until we decide
```
aghecenco, line 92: Since it's hardcoded, it would be better to declare it `const` somewhere.

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,263 @@
...
+        // to do something else.
+        assert_eq!(mmio_cfg.range.size(), 0x1000);
+
+        // Extra parameters have to be appended to the cmdline passed to the kernel because
+        // there's no active enumeration/discovery mechanism for virtio over MMIO. In the future,
+        // we might rely on a device tree representation instead.
+
+        // TODO: We add the "root=..." option everytime. Add an argument to specifiy whether
```
aghecenco, line 100: I think the `root` assignment should be specified in `BlockArgs`.

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,263 @@
...
+            .register_ioevent(
+                &ioeventfd,
+                &IoEventAddress::Mmio(
+                    self.mmio_cfg.range.base().0 + VIRTIO_MMIO_QUEUE_NOTIFY_OFFSET,
+                ),
+                0u32,
+            )
+            .map_err(Error::RegisterIoevent)?;
+
+        // TODO: Some changes required here to add read-only support as well.
```
aghecenco, line 164: These also need to go in `BlockArgs`.

coverage_config_x86_64.json
```diff
@@ -1,5 +1,5 @@
 {  
-  "coverage_score": 88.1,
+  "coverage_score": 74.3,
```
aghecenco, line 2: :broken_heart:

src/api/src/lib.rs
```diff
@@ -10,7 +10,7 @@ use std::convert::TryFrom;
 use std::result;
 
 use clap::{App, Arg};
-use vmm::{KernelConfig, MemoryConfig, VMMConfig, VcpuConfig};
+use vmm::{BlockConfig, KernelConfig, MemoryConfig, NetConfig, VMMConfig, VcpuConfig};
```
aghecenco, line 13: I'd remove `NetConfig` from this PR and save it for later.

aghecenco:
Functionality-wise looks good; I think it needs some more tinkering to customize the block args (option to specify which, if any, (root) device it corresponds to in the guest, plus RW permissions - I think both should be `Option`s and have defaults, to keep it simple), and quite some tinkering on tests. We'll have to figure out how to make block devices available in CI for integration tests.

src/vm-vcpu/src/vcpu/mod.rs
alexandruag: Sure, I'll add an issue if we merge the PR like this. This commit was needed for the VMM to actually terminate on reboot, or so it seemed when I cherry-picked it.

src/vm-vcpu/src/vcpu/mod.rs
alexandruag: @andreeaflorescu can you please help with a bit of context?

Cargo.toml
alexandruag: Will do after I finish answering the other comments here.

src/devices/src/virtio/block/device.rs
alexandruag: This was a sort-of lazy solution to the problem of converting the length of the MMIO region to a proper textual representation (altho, in retrospect, just using the raw number of bytes should work as well). Looks like this will be nicely fixed by using the addition proposed in https://github.com/rust-vmm/linux-loader/pull/53, and passing a `Cmdline` mutable handle instead of that `String`.

src/devices/src/virtio/block/device.rs
alexandruag: Will add.

src/devices/src/virtio/block/device.rs
alexandruag: Will add.

src/devices/src/virtio/block/device.rs
alexandruag: Hmm, this is not meant to be `x86_64` specific ... besides the current issue that prevents it from working properly on `aarch64`, what else looks x86 specific? Also, I'd keep the `ignore` so the test explicitly shows up as `ingored` as an additional signal/reminder that we need to fix it, but we can drop this if it's too verbose.

src/devices/src/virtio/block/mod.rs
alexandruag: Hmm, if I do that then the path might no longer be available, but I'll try it out thanks!

src/devices/src/virtio/block/queue_handler.rs
alexandruag: I don't think additional flags can be encountered for an `EventFd` :-s I'll add some logic to deregister the fd when we get an error.

src/devices/src/virtio/mod.rs
alexandruag: I tried to think about this as well, but didn't come up with a nicer option ('Params` or something like that is another, but not great either). Ultimately we can rename this as soon as we can think of something else.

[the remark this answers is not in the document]
alexandruag: The struct is already in the `virtio` module, so we can refer to it as `virtio::CommonArgs`. I think the name is quite self-explanatory, but if ppl think we get more clarity by changing it then sure.

src/devices/src/virtio/mod.rs
```diff
@@ -0,0 +1,150 @@
...
+/// Simple trait to model the operation of signalling the driver about used events
+/// for the specified queue.
+// TODO: Does this need renaming to be relevant for packed queues as well?
+pub trait SignalUsedQueue {
+    // TODO: Should this return an error? This failing is not really recoverable at the interface
+    // level so the expectation is the implementation handles that transparently somehow.
+    fn signal_used_queue(&self, index: u16);
+}
+
+/// Uses a single irqfd as the basis of signalling any queue (useful for the MMIO transport,
```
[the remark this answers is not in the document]
alexandruag: It's not necessarily for MMIO only. None of the names is great IMO :( We can think about this a bit more if it looks important at this point.

src/devices/src/virtio/mod.rs
alexandruag: sure, I'll rename it

src/devices/src/virtio/mod.rs
alexandruag: In the VMM case we can generate a `CommonArgs` struct backed by the vmm itself (which is needed anyway to create devices with the current model). This method is local to the devices crate to test things independently from the VMM if they ever get separated as crates. Otherwise, looks like we can just try to instantiate a VMM object as part of devices tests and generate `CommonArgs` based on that instead of mocking things.

src/vmm/Cargo.toml
```diff
@@ -5,19 +5,20 @@ authors = ["rust-vmm AWS maintainers <rust-vmm-maintainers@amazon.com>"]
...
 # vm-vcpu build, we're using a fixed revision.
 vm-device = { git = "https://github.com/rust-vmm/vm-device", rev = "5847f12" }
 
-vm-memory = { version = "0.3.0", features = ["backend-mmap"] }
-vm-superio = "0.1.1"
-vmm-sys-util = "0.6.1"
-
-libc = "0.2.76"
+devices = { path = "../devices" }
+vm-vcpu = { path = "../vm-vcpu" }
```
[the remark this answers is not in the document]
alexandruag: will do

coverage_config_x86_64.json
alexandruag: :(

src/api/src/lib.rs
alexandruag: It's not super clean, but the net PR will follow immediately after so it seemed more straightforward to do things like that.

src/vmm/src/lib.rs
alexandruag: sure

src/vmm/src/lib.rs
alexandruag: The current plan is to convert several parts of the current functionality to be re-usable for all virtio devices in the ref-vmm in the upcoming net PR.

src/vmm/src/lib.rs
alexandruag: sure

alexandruag (commented):
Thanks for the comments! I'll start implementing changes and will ping when things are ready for a second review.

src/devices/src/virtio/block/device.rs
andreeaflorescu: It is specific because:
- the mock is calling into vm.create_irq_chip (I left a suggestion where it is introduced about how this could become less specific)
- it is checking the side effect introduced by the kernel commandline update which will not be present on aarch64.

In other words, the only check that we can do on both platform from the entire test is this one:
```
let bus_range = ctx.mmio_mgr.mmio_device(range.base()).unwrap().0;
assert_eq!(bus_range.base(), range.base());
assert_eq!(bus_range.size(), range.size());
```

alexandruag: Cool, thanks for explaining! So, to add a bit of nuance, the current logic appears to be mostly for `x86` because the Arm specific stuff is not there yet to test, but when both are available, we should have platform specific-checks as well in here. I like that approach to avoid duplication of the common things, but we'll have a better perspective when Arm support starts getting added.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: This is what kvmtool does as well, didn't really do a dive deep.

src/devices/src/virtio/mod.rs
andreeaflorescu: Instantiating a VMM is not great either because it will make devices depend on vmm and that would be a cyclic dependency. This is why I was suggesting to just pass this as a parameter as a sort of okay compromise. It is sort of a nit up to the point where we actually add aarch64 support.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: We can update the default kernel command line to use "reboot=t" instead of "reboot=k" until we have a proper keyboard implementation that allows cleaner reboots.

src/devices/src/virtio/mod.rs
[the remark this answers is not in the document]
andreeaflorescu: You are not using the namespacing when using the CommonArgs outside of the virtio module which makes it a bit confusing.

In vmm/src/lib.rs it is used in the `add_block_device` function as follows:
```
        let common = CommonArgs {
            mem,
            vm_fd: self.vm.vm_fd(),
            event_mgr: &mut self.event_mgr,
            mmio_mgr: guard.deref_mut(),
            mmio_cfg,
            kernel_cmdline: &mut self.kernel_cfg.cmdline,
        };

        let args = BlockArgs {
            common,
            file_path: PathBuf::from(&cfg.path),
        };
```
At this point, the namespacing information is lost. We can also do the following in the vmm code:
```rust
use devices::virtio::{CommonArgs as VirtioCommonArgs, MmioConfig};
```

src/devices/src/virtio/block/queue_handler.rs
andreeaflorescu: I think events can contain at the same time EventSet::IN | EventSet::ERROR or EventSet::IN | EventSet::HUP.

src/devices/src/virtio/mod.rs
alexandruag: The only thing I'm wondering is whether passing that as a parameter doesn't just kick the can a bit forward, because someone still has to properly set that up. I'll implement any change, but if it's ok with you, I'd wait until we can actually test some Arm stuff as well.

src/vm-vcpu/src/vcpu/mod.rs
alexandruag: Sounds good ... I'll drop this commit then.

alexandruag (commented):
I've added changes to address open items from the questions. Regarding the CI, I think we're in a better position to do something about that once we add net support (coming right up) as well, but def open to other perspectives too. Would be great if you can take another look.

src/devices/src/virtio/block/mod.rs
lauralt: Ah sorry, didn't notice the `as_path()`.

Actually, `as_file()` works fine, i.e. `let mut file = tmp.as_file();`.

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,285 @@
...
+            common: common_args,
+            file_path: tmp.as_path().to_path_buf(),
+            read_only: false,
+            root_device: true,
+        };
+
+        let _block = Block::new(args).unwrap();
+
+        // Some quick sanity checks. Most of the functionality around the device should be
+        // exercised/validated via integration tests.
```
lauralt, line 268: Hmm, can't we at least check that the device features from `virtio_cfg` are the expected ones and maybe check other scenarios as well (e.g. read_only -> true, root_device -> false)?

src/api/src/lib.rs
```diff
@@ -89,6 +102,20 @@ impl CLI {
                 eprintln!("{}", help_msg);
                 format!("{}", e)
             })?,
+            block_config: match matches.value_of("block") {
+                Some(value) => Some(BlockConfig::try_from(value.to_string()).map_err(|e| {
+                    eprintln!("{}", help_msg);
+                    format!("{}", e)
+                })?),
+                None => None,
```
lauralt, line 110: Wouldn't some `map` work here to avoid the `None => None` line? :(

src/vmm/src/lib.rs
```diff
@@ -157,8 +165,12 @@ impl TryFrom<VMMConfig> for VMM {
             kernel_cfg: config.kernel_config.clone(),
         };
 
-        vmm.add_serial_console()?;
         vmm.create_vcpus(&config.vcpu_config)?;
+        vmm.add_serial_console()?;
+
+        if let Some(block_cfg) = config.block_config.as_ref() {
+            vmm.add_block_device(block_cfg)?;
```
andreeaflorescu, line 172: Can you also add a test for `TryFrom` where you also pass a block device configuration?

src/vmm/src/lib.rs
```diff
@@ -301,6 +313,36 @@ impl VMM {
...
+        };
+
+        let args = BlockArgs {
+            common,
+            file_path: PathBuf::from(&cfg.path),
+            read_only: false,
+            root_device: true,
+        };
+
+        // We can also hold this somewhere if we need to keep the handle for later.
```
andreeaflorescu, line 340: Even though we are no longer interacting with the block device after we create it, it would be nice to keep the handle in the VMM so that we can write more meaningful tests.

src/vmm/src/lib.rs
```diff
@@ -301,6 +313,36 @@ impl VMM {
         Ok(())
     }
 
+    fn add_block_device(&mut self, cfg: &BlockConfig) -> Result<()> {
```
andreeaflorescu, line 316: Can you add unit tests for this code? It should be relatively straightforward to use the `mock_vmm`, and check the valid scenario: path is valid and exists, and at least one invalid scenario so that we check that the appropriate value is returned.

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,285 @@
...
+    fn test_device() {
+        let mut ctx = CommonArgsMock::new();
+        let tmp = TempFile::new().unwrap();
+
+        let common_args = ctx.args();
+        let args = BlockArgs {
+            common: common_args,
+            file_path: tmp.as_path().to_path_buf(),
+            read_only: false,
+            root_device: true,
```
andreeaflorescu, line 262: Can we also add a test for read_only=true and root_device=false? We should check the resulting command line and the device features.

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,285 @@
...
+    type E = Error;
+
+    fn activate(&mut self) -> Result<()> {
+        if self.virtio_cfg.device_activated {
+            return Err(Error::AlreadyActivated);
+        }
+
+        if !self.queues_valid() {
+            return Err(Error::QueuesNotValid);
+        }
```
andreeaflorescu, line 153: While successful activation is a bit overkill to test, we can easily test these two error scenarios (and the device_type function) as follows:
```rust
        let block = Block::new(args).unwrap();
        assert_eq!(block.lock().unwrap().device_type(), BLOCK_DEVICE_ID);

        assert!(matches!(
            block.lock().unwrap().activate(),
            Err(Error::QueuesNotValid)
        ));

        block.lock().unwrap().virtio_config_mut().device_activated = true;
        assert_eq!(block.lock().unwrap().virtio_config().device_activated, true);
        assert!(matches!(
            block.lock().unwrap().activate(),
            Err(Error::AlreadyActivated)
        ));
```

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,285 @@
...
+            queue: self.virtio_cfg.queues[0].clone(),
+            disk,
+        };
+
+        let handler = Arc::new(Mutex::new(QueueHandler { inner, ioeventfd }));
+
+        // Register the queue handler with the `EventManager`. We could record the `sub_id`
+        // (and/or keep a handler clone) for further interaction (i.e. to remove the subscriber at
+        // a later time, retrieve state, etc).
+        let _sub_id = self
```
andreeaflorescu, line 211: Should we remember this subscriber id and actually use it in the process function to de-register the device when we encounter an error?

src/vmm/src/lib.rs
alexandruag: You mean a new test, or is there some existing test somewhere?

src/vmm/src/lib.rs
alexandruag: Added a `block_devices` member to `VMM`. Is that similar to what you had in mind?

src/vmm/src/lib.rs
alexandruag: I'd like to add this kind of testing in the following PR where I do some refactoring of the common bits of code between devices (i.e. block and net at that point)

alexandruag (commented):
Addressed more comments and added some more of the suggested tests. There's a bit of a hack in the last commit regarding the use of `Cmdline` from `kernel_loader`. The devices now expect a `&mut Cmdline` reference at creation time (instead of a `&mut String`), but the VMM uses strings (and variations thereof) instead of using `Cmdline`. Let's decide what the path forward is here. I'll also start working on some basic integration tests for the CI.

src/devices/src/virtio/block/device.rs
```diff
@@ -0,0 +1,285 @@
...
+        // pointers such as mutex guards here.
+        B: DerefMut,
+        B::Target: MmioManager<D = Arc<dyn DeviceMmio + Send + Sync>>,
+    {
+        // The queue handling logic for this device uses the buffers in order, so we enable the
+        // corresponding feature as well.
+
+        // TODO: We don't offer flush (1 << VIRTIO_BLK_F_FLUSH), so writes are faster overall but
+        // not guaranteed to be persisted to physical storage in the event of something like power
+        // loss. Add this as a configuration option?
```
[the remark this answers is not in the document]
alexandruag: my bad; done

src/devices/src/virtio/block/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+}
+
+pub type Result<T> = std::result::Result<T, Error>;
+
+// TODO: Add a helper abstraction to rust-vmm for building the device configuration space.
+// The one we build below for the block device contains the minimally required `capacity` member,
+// but other fields can be present as well depending on the negotiated features.
+fn build_config_space<P: AsRef<Path>>(path: P) -> Result<Vec<u8>> {
+    // TODO: right now, the file size is computed by the StdioBackend as well. Maybe we should
+    // create the backend as early as possible, and get the size information from there.
```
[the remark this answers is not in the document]
alexandruag: That's interesting to consider. After/if we change the interface of `StdioBackend` like that, we can propagate the rest of the changes here as well.

src/devices/src/virtio/block/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+        .map_err(Error::Seek)?;
+    // If the file size is actually not a multiple of sector size, then data at the very end
+    // will be ignored.
+    let num_sectors = file_size >> SECTOR_SHIFT;
+    // This has to be in little endian btw.
+    Ok(num_sectors.to_le_bytes().to_vec())
+}
+
+// Arguments required when building a block device.
+// TODO: Add read-only operation support as a quick next step.
```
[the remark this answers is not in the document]
alexandruag: my bad; done

src/devices/src/virtio/block/mod.rs
```diff
@@ -0,0 +1,123 @@
...
+        {
+            let config_space = build_config_space(tmp.as_path()).unwrap();
+
+            // The config space can only be populated with the `capacity` field for now.
+            assert_eq!(config_space.len(), size_of::<u64>());
+
+            // The `try_into` attempts to convert the byte slice into a `[u8; 8]` array, as required
+            // by `u64::from_le_bytes`.
+            let capacity = u64::from_le_bytes(config_space[..8].try_into().unwrap());
+            assert_eq!(capacity, num_sectors);
```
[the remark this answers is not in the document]
alexandruag: It is right now, but the config space for a block device can be larger than 8 bytes, depending on the presence of various features (more details [here](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-2440004)). I would like to keep this as is, but will change it if there's a strong preference for the alternative.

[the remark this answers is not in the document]
lauralt: `assert_eq!(config_space[..8], num_sectors.to_le_bytes());` doesn't do the trick?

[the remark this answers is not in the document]
alexandruag: It does, thanks!

alexandruag (commented):
Addressed the most recent comments as well, and added a simple CI test for the block device.

src/vmm/src/lib.rs
andreeaflorescu: Yes.

andreeaflorescu: Actually, I've looked a bit more at this and it looks like we only support one device right now. Moreover, calling the `add_block_device` twice would yield an error because the `MmioConfig` is hard coded. How about we actually use Option<Block> instead of a vector for now? And also maybe check if the self.block is None, otherwise return an error?

src/vmm/src/lib.rs
```diff
@@ -303,6 +329,42 @@ impl VMM {
         Ok(())
     }
 
+    fn add_block_device(
```
andreeaflorescu, line 332: This can also be unit tested and we can check for side effects as follows:
```rust
    #[test]
    fn test_add_block() {
        let vmm_config = default_vmm_config();
        let mut vmm = mock_vmm(vmm_config);

        let tempfile = TempFile::new().unwrap();
        let block_config = BlockConfig {
            path: tempfile.as_path().to_path_buf()
        };
        let mut cmdline = cmdline::Cmdline::new(4096);
        assert!(vmm.add_block_device(&block_config, &mut cmdline).is_ok());
        assert_eq!(vmm.block_devices.len(), 1);
        assert!(cmdline.as_str().contains("virtio"));

        let invalid_block_config = BlockConfig {
            // Let's create the tempfile directly here so that it gets out of scope immediately
            // and delete the underlying file.
            path: TempFile::new().unwrap().as_path().to_path_buf()
        };
        let mut cmdline = cmdline::Cmdline::new(4096);
        let err = vmm.add_block_device(&invalid_block_config, &mut cmdline).unwrap_err();
        assert!(matches!(err, Error::Block(block::Error::OpenFile(io_err)) if io_err.kind() == ErrorKind::NotFound));

        // The current implementation of the VMM does not allow more than one block device
        // as we are hard coding the `MmioConfig`.
        let mut cmdline = cmdline::Cmdline::new(4096);
        assert!(vmm.add_block_device(&block_config, &mut cmdline).is_err());
    }
```

src/vmm/src/lib.rs
```diff
@@ -157,10 +168,25 @@ impl TryFrom<VMMConfig> for VMM {
...
+        // apparently does not explicitly use `Cmdline` structs. Will discuss and fix this
+        // somehow ASAP.
+        let mut cmdline = cmdline::Cmdline::new(4096);
+
+        if let Some(block_cfg) = config.block_config.as_ref() {
+            vmm.add_block_device(block_cfg, &mut cmdline)?;
+        }
+
+        vmm.kernel_cfg.cmdline.push_str(" ");
+        vmm.kernel_cfg.cmdline.push_str(cmdline.as_str());
```
andreeaflorescu, line 189: This is a bit ugly because if we have no additional devices there is no need to add the extra space to the cmdline. Can we for the time being move the cmdline changes in the previous if block?

src/vmm/src/lib.rs
```diff
@@ -37,14 +40,17 @@ use vm_vcpu::vcpu::{cpuid::filter_cpuid, VcpuState};
 use vm_vcpu::vm::{self, KvmVm, VmState};
 use vmm_sys_util::{eventfd::EventFd, terminal::Terminal};
 
+use devices::virtio::block::{self, BlockArgs};
+use devices::virtio::{CommonArgs, MmioConfig};
```
andreeaflorescu, line 44: nit: can you import CommonArgs as VirtioCommonArgs?

tests/test_run_reference_vmm.py
```diff
@@ -66,34 +82,48 @@ def test_reference_vmm(kernel):
...
-    """Start the reference VMM and verify that it works."""
+    """Start the reference VMM with a block device and verify that it works."""
 
-    # TODO: that's actually a lie, this test is a placeholder that just
-    # verifies that the prerequisites are in place.
+    vmm_process = start_vmm_process(kernel, disk)
 
-    kernel_path = resource_path("kernel", kernel)
-    disk_path = resource_path("disk", disk)
+    expect_string(vmm_process, "Ubuntu 20.04 LTS ubuntu-rust-vmm ttyS0")
```
andreeaflorescu, line 127: Should we also run the `lsblk --json` and check fields such as:
```"ro":false && "name":"vda"```

File: tests/test_run_reference_vmm.py [hunk 36]
```diff
@@ -8,9 +8,14 @@
 
 
 KERNELS_INITRAMFS = ["vmlinux-hello-busybox", "bzimage-hello-busybox"]
+
+"""
+Temporarily removed the ("bzimage-focal", "rootfs.ext4") pair from the
```
lauralt, line 13: Should we open an issue?

tests/test_run_reference_vmm.py
alexandruag: Would be great if we can merge the current iteration of the test, and improve it later (created #47).

lauralt (dismissed):

src/vmm/src/config.rs
```diff
@@ -26,6 +30,8 @@ impl fmt::Display for ConversionError {
             ParseKernel(ref s) => write!(f, "Invalid input for kernel: {}", s),
             ParseMemory(ref s) => write!(f, "Invalid input for memory: {}", s),
             ParseVcpus(ref s) => write!(f, "Invalid input for vCPUs: {}", s),
+            ParseNetwork(ref s) => write!(f, "Invalid input for network: {}", s),
+            ParseBlock(ref s) => write!(f, "Invalid input for network: {}", s),
```
lauralt, line 34:
```suggestion
            ParseBlock(ref s) => write!(f, "Invalid input for block: {}", s),
```

lauralt (approved):

andreeaflorescu (approved):

src/vmm/src/config.rs
alexandruag: Thanks, will include this fix in the following PR.

PR #42, opened 2020-12-04
Build disk images in CI

aghecenco:


aghecenco:
Needs https://github.com/rust-vmm/rust-vmm-container/pull/34 and a new `rust-vmm-container` image to be clean.

.buildkite/deps-pipeline.yml
```diff
@@ -1,10 +1,10 @@
 steps:
   - label: "build-deps-x86"
     commands:
-      - resources/kernel/make_kernel_image.sh -f elf -k vmlinux-hello-busybox -w /tmp/vmlinux_busybox -j 2
-      - resources/kernel/make_kernel_image.sh -f elf -k vmlinux-hello-busybox -w /tmp/vmlinux_busybox -j 2 -h
-      - resources/kernel/make_kernel_image.sh -f bzimage -k bzimage-hello-busybox -w /tmp/vmlinux_busybox -j 2
-      - resources/kernel/make_kernel_image.sh -f bzimage -k bzimage-hello-busybox -w /tmp/vmlinux_busybox -j 2 -h
+      - resources/kernel/make_kernel_initramfs_image.sh -f elf -k vmlinux-hello-busybox -w /tmp/vmlinux_busybox -j 2
```
alexandruag, line 4: Is this script busybox-specific? If so, `make_busybox_image` might be a more accurate name/description.

resources/kernel/make_kernel.sh
```diff
@@ -149,20 +149,21 @@ kernel_binary() {
...
 #       format              \
+#       [target]            \
 #       [num_cpus_build]    \
 #       [/path/to/kernel/destination]
 make_kernel() {
     kernel_dir="$1"
     format="$2"
-    nprocs="$3"
-    dst="$4"
+    target="$3"
```
alexandruag, line 158: Hmm, why is this added as an extra param versus how things were before?

resources/kernel/make_kernel_image_deb.sh
```diff
@@ -0,0 +1,127 @@
...
+# of building the entire kernel over and over again.
+# If, however, you want to build them from scratch, set the `MAKEDEB` env var.
+mkdir -p "$kernel_dir/deb"
+if [ -n "$MAKEDEB" ]; then
+    echo "Building deb packages..."
+    make_kernel "$kernel_dir" "$KERNEL_FMT" "deb-pkg" "$MAKEPROCS"
+    cp "$kernel_dir"/../*.deb "$kernel_dir/deb"
+else
+    echo "Downloading deb packages..."
+    DEB_URL="http://security.ubuntu.com/ubuntu/pool/main/l/linux-hwe-5.4"
```
alexandruag, line 106: Hmmm, what are the deb packages for? It seems there are variables for things such as the kernel version, but then we have these links hard-coded here.

.buildkite/deps-pipeline.yml
aghecenco: Yup, it builds kernels witth busybox initramfs'es. I'll re-rename it.

resources/kernel/make_kernel.sh
aghecenco: Because this function can be recycled to build `.deb`s with `make deb-pkg`. So the `target` is no longer a variable of the kernel format.

resources/kernel/make_kernel_image_deb.sh
aghecenco: It's the kernel image and modules. Even though we don't use the kernel shipped with the `linux-image` deb, Ubuntu still needs a bunch of modules in the `linux-modules` deb (and `System.map` for some reason) to work properly.

aghecenco: As for the kernel version: we don't anywhere to download the *exact* kernel Ubuntu 20.04 ships with, so we instead build one with the same major.minor and thankfully it works fine. Turns out, you can't find *everything* on the Internet :(

alexandruag: Hmm, aren't the modules and System.map available after building the kernel from sources as well? Also, does it make sense to use a kernel config where the required things are compiled as part of the image so we don't need to load modules?

aghecenco: I think they can be built with other `make` recipes, but it takes way longer than it does to download them, and I'm not sure how to all the pieces come together to make a full Ubuntu. I went with this 2-step solution because it did the trick, and doesn't require multiple (re)compilations, but if you know a quicker/simpler way to get the working OS, we should go with it.

alexandruag: I don't TBH, but the current one was/is a bit hard to understand and I was wondering if we can simplify things. I guess one way forward here is to go as things are until we have the initial devices merged and exercised by some tests as part of the CI etc, and then take a better look at the system itself, and at the very least add some more documentation if no other changes are required or important for clarity and reuse. What do other folks think here?

docs/getting-started.md
```diff
@@ -67,30 +68,79 @@ directory (default: `target/${toolchain}/${mode}`, where mode can be `debug` or
...
+  directory. Root privileges are needed to create device nodes.
 
-  Run `make_kernel_image.sh` with no arguments to see the help. For more
-  usage examples, see the [Buildkite pipeline](../.buildkite/deps-pipeline.yml)
-  that calls this script as part of the CI.
+  Run `make_kernel_busybox_image.sh` with no arguments to see the help.
+
+- [`make_kernel_image_deb.sh`](../resources/kernel/make_kernel_image_deb.sh)
+  builds an ELF or bzImage kernel compatible with Ubuntu 20.04 from a
+  stripped-down [kernel config](microvm-kernel-5.4-x86_64.config), as well as
```
lauralt, line 91: This path should be updated.

resources/disk/make_rootfs.sh
```diff
@@ -0,0 +1,125 @@
...
+
+Options:
+  -d debdir         Directory containing .deb packages for the Linux image.
+  -w workdir        Working directory for the kernel build.
+  -o diskfile       Name of the resulting disk file.
+  -c                Clean up the working directory after the build.
+"
+export USAGE
+
+while getopts ":ch:d:w:o:" opt; do
```
lauralt, line 44: Looks like we don't have a `h` option anymore and can be removed from here.

resources/kernel/make_kernel_image_deb.sh
lauralt: We can go with how things are right now (the documentation is pretty clear) and, like you've said, think about simplifications after devices are tested by the CI. Also there is some duplicate code in the kernel scripts that could probably be avoided, but for now I'm okay with the PR.

tests/test_run_reference_vmm.py
```diff
@@ -19,20 +23,24 @@ def process_exists(pid):
...
+@pytest.mark.parametrize("kernel", KERNELS_INITRAMFS)
 def test_reference_vmm(kernel):
-    """Start the reference VMM and trust that it works."""
+    """Start the reference VMM and verify that it works."""
 
     # Memory config
     mem_size_mib = 1024
 
     # Kernel config
     cmdline = "console=ttyS0 i8042.nokbd reboot=k panic=1 pci=off"
```
lauralt, line 42: nit: can we update the `cmdline` with the suggestion from [here](https://github.com/rust-vmm/vmm-reference/pull/41#discussion_r534318525)?

alexandruag (commented):
Looks like we can merge this after addressing the remaining comments.

lauralt (approved):

alexandruag (approved):

Issue #44, opened 2020-12-08
make_busybox.sh doesn't work


qhjindev:
When I started building the kernel according the getting-started.md, I executed the cmd under ubuntu to build an elf image:
  ./make_busybox.sh
But I can't see anything created and no output from the command line.

andreeaflorescu:
We need to update the docs. The scripts for creating a busybox were updated. You should use `make_kernel_image.sh` instead.

There are a few usage examples here: https://github.com/rust-vmm/vmm-reference/blob/master/.buildkite/deps-pipeline.yml#L3

Let us know if you need additional help.

PR #45, opened 2020-12-08
Docs: update getting started guide about kernel generation scripts

aghecenco:
#44

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -67,19 +67,25 @@ directory (default: `target/${toolchain}/${mode}`, where mode can be `debug` or
 To build a kernel for the reference VMM to boot, check out the scripts in
 [resources/kernel](../resources/kernel).
 
-- [`make_busybox.sh`](../resources/kernel/make_busybox.sh) builds an ELF image
-  with a baked-in initramfs running [Busybox](https://busybox.net/). It uses a
-  stripped-down
+- [`make_kernel_image.sh`](../resources/kernel/make_kernel_image.sh) builds an
+  ELF or bzImage kernel with a baked-in initramfs running
+  [Busybox](https://busybox.net/). It uses a stripped-down
   [kernel config](../resources/kernel/microvm-kernel-initramfs-hello-x86_64.config)
   and a statically linked [config](../resources/kernel/busybox_static_config)
   for the Busybox initramfs.
 
+  For example:
+
   ```bash
-  ./make_busybox.sh
+  ./make_kernel_image.sh -f elf -k vmlinux-hello-busybox -w /tmp/kernel
   ```
 
-  This produces a binary image called `vmlinux-hello-busybox` in the
-  `resources/kernel` directory.
+  produces a binary image called `vmlinux-hello-busybox` in the `/tmp/kernel`
+  directory.
+
+  Run `make_kernel_image.sh` with no arguments to see the help. For more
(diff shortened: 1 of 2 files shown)


docs/getting-started.md
```diff
@@ -67,19 +67,25 @@ directory (default: `target/${toolchain}/${mode}`, where mode can be `debug` or
...
+  [Busybox](https://busybox.net/). It uses a stripped-down
   [kernel config](../resources/kernel/microvm-kernel-initramfs-hello-x86_64.config)
   and a statically linked [config](../resources/kernel/busybox_static_config)
   for the Busybox initramfs.
 
+  For example:
+
   ```bash
-  ./make_busybox.sh
+  ./make_kernel_image -f elf -k vmlinux-hello-busybox -w /tmp/kernel
```
lauralt, line 80: If the script name contains `.sh` in its name, do we expect to be able to run the script without specifying the whole name (without `.sh`)? Honest question, I googled it and no clear answer. Also, from what it looks like, a god practice is to not include the extension in the name, there's a thread here about this: https://askubuntu.com/questions/764121/sh-specifying-extension.
Anyway, to avoid confusions I would use `make_kernel_image.sh` everywhere or rename it to not include the extension.

aghecenco: Fixed!

lauralt (approved):

docs/getting-started.md
```diff
@@ -67,19 +67,25 @@ directory (default: `target/${toolchain}/${mode}`, where mode can be `debug` or
 To build a kernel for the reference VMM to boot, check out the scripts in
 [resources/kernel](../resources/kernel).
 
-- [`make_busybox.sh`](../resources/kernel/make_busybox.sh) builds an ELF image
-  with a baked-in initramfs running [Busybox](https://busybox.net/). It uses a
-  stripped-down
+- [`make_kernel_image.sh`](../resources/kernel/make_kernel_image.sh) builds an
```
alexandruag, line 70: Is this script the one that's ultimately renamed to `make_kernel_busybox_image` in #42? Also I guess something like the original `make_busybox` is the shortest and clearest name (especially with the additional documentation available), but if you prefer another name that's fine as well.

aghecenco: Yup, it's the same script. There's already a `make_busybox` script that builds just the busybox image. This script calls into it as part of the larger kernel+initramfs build. I'm (obviously) not good with names - do you think the helper scripts should be renamed as well, so that the "user-facing" scripts keep concise names?

alexandruag: Would be nice for the scripts that are user-facing to have the nicer names if you already have an approach here mind. We can have a second look at the whole thing when rebasing #42 on top of this (if that's how we are going to proceed). Meanwhile, just wondering, I see Laura's comment above about removing the `.sh` extension, but then it seems it's still present when looking at the actual commit :-s Am I missing/looking at something else here? :-s

lauralt: It wasn't a hard requirement or something like that for this particular PR. I mostly wanted to not have both `make_kernel_image.sh` and `make_kernel_image` in docs, scripts etc. If you are ok with renaming it, we can do that either in this PR or #42.

alexandruag (approved):
Ok, let's keep things like this for now and see how they look in the context of #42 and beyond.

PR #48, opened 2020-12-14
deps pipeline: add wait step

aghecenco:
The rootfs creation step depends on the kernel build step, so it should wait for it to finish.


aghecenco:
The fix was included in #41. Closing.

PR #49, opened 2020-12-15
Add net device support

alexandruag:
This PR refactors some of the common virtio device code to reduce the amount of required boilerplate, and then implements virtio net device support. The `tap` functionality and associated logic is taken from Firecracker for now, until we'll have a similar component available in rust-vmm or someplace else. More tests have to be added, and we're going to circle back to that after the initial version of the vsock support PR is ready as well.

Helpful steps for testing net device support:
- A `tap` device is needed on the host. One can be created and configured with something like (please note the example might conflict with the existing host configuration if there's overlap in terms of IP addresses/subnets):
```
sudo ip tuntap add dev vmtap100 mode tap
sudo ip addr add 192.168.241.1/24 dev vmtap100
sudo ip link set vmtap100 up
```
- Start the reference vmm with the `--net` parameter, i.e. `cargo run ... -- ... --net tap=vmtap100`

- A network interface should be present inside the guest (usually `eth0`, but the exact naming may vary). We need to configure this interface as well, for example by running (inside the guest):
```
ip a add 192.168.241.2/24 dev eth0
ip link set eth0 up
``` 

Please note the interface in the guest and the `tap` device from the host must be configured as part of the same subnet ( `192.168.241.0/24` in the previous examples) to communicate directly.

### DIFF
--- a/src/devices/src/virtio/net/device.rs
+++ b/src/devices/src/virtio/net/device.rs
@@ -1,0 +1,150 @@
+// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use std::borrow::{Borrow, BorrowMut};
+use std::ops::DerefMut;
+use std::sync::{Arc, Mutex};
+
+use vm_device::bus::MmioAddress;
+use vm_device::device_manager::MmioManager;
+use vm_device::{DeviceMmio, MutDeviceMmio};
+use vm_memory::GuestAddressSpace;
+use vm_virtio::device::{VirtioConfig, VirtioDeviceActions, VirtioDeviceType, VirtioMmioDevice};
+use vm_virtio::Queue;
+
+use crate::virtio::features::{VIRTIO_F_IN_ORDER, VIRTIO_F_RING_EVENT_IDX, VIRTIO_F_VERSION_1};
+use crate::virtio::net::features::*;
+use crate::virtio::net::{Error, NetArgs, Result, NET_DEVICE_ID, VIRTIO_NET_HDR_SIZE};
+use crate::virtio::{CommonConfig, Env, SingleFdSignalQueue, QUEUE_MAX_SIZE};
+
+use super::bindings;
+use super::queue_handler::QueueHandler;
+use super::simple_handler::SimpleHandler;
+use super::tap::Tap;
+
+pub struct Net<M: GuestAddressSpace> {
+    cfg: CommonConfig<M>,
+    tap_name: String,
+}
+
+impl<M> Net<M>
+where
+    M: GuestAddressSpace + Clone + Send + 'static,
+{
+    pub fn new<B>(env: &mut Env<M, B>, args: &NetArgs) -> Result<Arc<Mutex<Self>>>
+    where
+        // We're using this (more convoluted) bound so we can pass both references and smart
+        // pointers such as mutex guards here.
+        B: DerefMut,
+        B::Target: MmioManager<D = Arc<dyn DeviceMmio + Send + Sync>>,
+    {
+        let device_features = (1 << VIRTIO_F_VERSION_1)
+            | (1 << VIRTIO_F_RING_EVENT_IDX)
+            | (1 << VIRTIO_F_IN_ORDER)
+            | (1 << VIRTIO_NET_F_CSUM)
+            | (1 << VIRTIO_NET_F_GUEST_CSUM)
+            | (1 << VIRTIO_NET_F_GUEST_TSO4)
+            | (1 << VIRTIO_NET_F_GUEST_TSO6)
+            | (1 << VIRTIO_NET_F_GUEST_UFO)
+            | (1 << VIRTIO_NET_F_HOST_TSO4)
+            | (1 << VIRTIO_NET_F_HOST_TSO6)
+            | (1 << VIRTIO_NET_F_HOST_UFO);
+
+        // An rx/tx queue pair.
+        let queues = vec![Queue::new(env.mem.clone(), QUEUE_MAX_SIZE); 2];
+        // TODO: We'll need a minimal config space to support setting an explicit MAC addr
+        // on the guest interface at least. We use an empty one for now.
+        let config_space = Vec::new();
+        let virtio_cfg = VirtioConfig::new(device_features, queues, config_space);
+
+        let common_cfg = CommonConfig::new(virtio_cfg, env).map_err(Error::Virtio)?;
+
+        let net = Arc::new(Mutex::new(Net {
+            cfg: common_cfg,
+            tap_name: args.tap_name.clone(),
+        }));
+
+        env.register_mmio_device(net.clone())
+            .map_err(Error::Virtio)?;
+
+        Ok(net)
+    }
+}
+
+impl<M: GuestAddressSpace + Clone + Send + 'static> VirtioDeviceType for Net<M> {
+    fn device_type(&self) -> u32 {
+        NET_DEVICE_ID
+    }
+}
+
+impl<M: GuestAddressSpace + Clone + Send + 'static> Borrow<VirtioConfig<M>> for Net<M> {
+    fn borrow(&self) -> &VirtioConfig<M> {
+        &self.cfg.virtio
+    }
+}
+
+impl<M: GuestAddressSpace + Clone + Send + 'static> BorrowMut<VirtioConfig<M>> for Net<M> {
+    fn borrow_mut(&mut self) -> &mut VirtioConfig<M> {
+        &mut self.cfg.virtio
+    }
+}
+
+impl<M: GuestAddressSpace + Clone + Send + 'static> VirtioDeviceActions for Net<M> {
+    type E = Error;
+
+    fn activate(&mut self) -> Result<()> {
+        let rxq = self.cfg.virtio.queues[0].clone();
+        let txq = self.cfg.virtio.queues[1].clone();
+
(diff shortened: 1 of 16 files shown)


src/devices/src/virtio/net/device.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+impl<M: GuestAddressSpace + Clone + Send + 'static> VirtioDeviceActions for Net<M> {
+    type E = Error;
+
+    fn activate(&mut self) -> Result<()> {
+        let rxq = self.cfg.virtio.queues[0].clone();
+        let txq = self.cfg.virtio.queues[1].clone();
+
+        let tap = Tap::open_named(self.tap_name.as_str()).map_err(Error::Tap)?;
```
aghecenco, line 97: Doesn't this also creates the device if it isn't already present? The PR description mentions how to create tap devices in order for the tests to run, but shouldn't they be created anyway assuming that the user has `CAP_NET_ADMIN`?

src/devices/src/virtio/net/queue_handler.rs
```diff
@@ -0,0 +1,95 @@
...
+
+impl<M: GuestAddressSpace> QueueHandler<M> {
+    // Helper method that receives an error message to be logged and the `ops` handle
+    // which is used to unregister all events.
+    fn handle_error<S: AsRef<str>>(&self, s: S, ops: &mut EventOps) {
+        error!("{}", s.as_ref());
+        ops.remove(Events::empty(&self.rx_ioevent))
+            .expect("Failed to remove rx ioevent");
+        ops.remove(Events::empty(&self.tx_ioevent))
+            .expect("Failed to remove rx ioevent");
```
aghecenco, line 32:
```suggestion
            .expect("Failed to remove tx ioevent");
```

aghecenco: I think it would be better to `ops.remove` them in a row, keep track of the results, and panic after all the `remove`s have been attempted so that events don't incorrectly remain in the event manager in case of a quick death.

That being said, I'm not sure about panicking in library code. There's not much we can do about it yet but if you go ahead with the `expect`s, please add an issue to track removing them once we have metrics.

src/devices/src/virtio/net/tap.rs
```diff
@@ -0,0 +1,456 @@
+// Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
aghecenco, line 1: I really prefer the `crosvm` solution of a `TapT` trait, which is implemented differently in tests. That way, you can unit test your code without `CAP_NET_ADMIN`.

src/devices/src/virtio/net/simple_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+
+        self.rxbuf_current = 0;
+
+        Ok(true)
+    }
+
+    pub fn process_tap(&mut self) -> result::Result<(), Error> {
+        loop {
+            if self.rxbuf_current == 0 {
+                match self.tap.read(&mut self.rxbuf) {
```
lauralt, line 111: I'm not quite sure what happens if we read some data from tap here, but don't have a descriptor chain available for writing to its buffer. Is that data lost? Should we do this `read` only if we're sure there is at least one descriptor chain?

src/devices/src/virtio/net/simple_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+        let buf = &mut self.rxbuf[..num_bytes];
+
+        while let Some(desc) = chain.next() {
+            let left = buf.len() - count;
+
+            if left == 0 {
+                break;
+            }
+
+            let len = cmp::min(left, desc.len() as usize);
```
lauralt, line 87: What happens if `len` field from the descriptor is always 0 and we don't write any data to memory (count = 0 at the end of this loop)? Does it still make sense to add an entry in the used ring when no data was actually transferred?

src/devices/src/virtio/net/simple_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    pub fn process_txq(&mut self) -> result::Result<(), Error> {
+        loop {
+            self.txq.disable_notification()?;
+
+            while let Some(mut chain) = self.txq.iter()?.next() {
+                let len = self.send_frame_from_chain(&mut chain)?;
+
+                self.txq.add_used(chain.head_index(), len)?;
```
lauralt, line 171: `len` should be replaced with 0 I think.

src/devices/src/virtio/net/simple_handler.rs
```diff
@@ -0,0 +1,188 @@
...
+
+            chain
+                .memory()
+                .read_slice(&mut self.txbuf[count..count + len], desc.addr())
+                .map_err(Error::GuestMemory)?;
+
+            count += len;
+        }
+
+        self.tap.write(&self.txbuf[..count]).map_err(Error::Tap)?;
```
lauralt, line 159: How should we treat the case when less than count bytes could've been written to the tap device? I guess we can add at least a (TODO) comment about this :-?.

src/vmm/src/lib.rs
```diff
@@ -162,27 +165,66 @@ impl TryFrom<VMMConfig> for VMM {
...
         vmm.add_serial_console()?;
 
+        // Adding the virtio devices. We'll come up with a cleaner abstraction for `Env`.
+
+        let mem = Arc::new(vmm.guest_memory.clone());
+
         // We transiently define a mut `Cmdline` object here to pass for device creation
         // (devices expect a `&mut Cmdline` to leverage the newly added virtio_mmio
         // functionality), until we figure out how this fits with the rest of the vmm, which
         // apparently does not explicitly use `Cmdline` structs. Will discuss and fix this
```
lauralt, line 188: It would be nice if we could use `Cmdline` everywhere (hopefully, we will avoid this way adding spaces in strings like [here](https://github.com/rust-vmm/vmm-reference/blob/master/src/vmm/src/lib.rs#L312), but didn't look very much into how is `Cmdline` implemented).

lauralt (commented):
Looks nice, left a few comments, most of them related to rx/tx processing.

alexandruag (commented):
Hi all, I've updated the PR, answered the previous comments, and opened some new issues. Would be useful to merge the current iteration, so we can add some more consolidation logic and device simplification on top of these changes. Here are some of the current concerns:

- The unit tests for the `Tap` struct taken from Firecracker/crosvm require elevated privileges to run, as you have already noticed. Given that we'd like `Tap` to be a separate abstraction (#94), should we just consider the current implementation essentially immutable after we've brought it over and not include those tests as well, or should we temporarily run the tests with more privileges (if they are not so already)?

- Not sure how creating and tearing down TAP interfaces should exist within the current integration testing framework (as it's required for integration testing). Does anyone have a good starting idea here? Should we create an issue and solve that separately going forward?

alexandruag:
Looks like the tap tests run ok on the CI system; in that case the current disadvantage is that ppl running locally have to explicitly provide the required privileges so they can pass. I lowered the coverage due to not having sufficient unit tests as part of the net functionality (the queue handlers in particular). Are you ok with tracking that as a separate issue while things start falling into place more within `vm-virtio` (and some of the device logic gets consolidated)?

src/devices/src/virtio/net/device.rs
```diff
@@ -0,0 +1,147 @@
...
+
+        // Set offload flags to match the relevant virtio features of the device (for now,
+        // statically set in the constructor.
+        tap.set_offload(
+            bindings::TUN_F_CSUM
+                | bindings::TUN_F_UFO
+                | bindings::TUN_F_TSO4
+                | bindings::TUN_F_TSO6,
```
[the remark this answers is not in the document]
alexandruag: As you recently pointed out, I forgot to add `VIRTIO_NET_F_GUEST_TSO6` to the feature list as well. I've done so, and will update the next time I push changes to the PR.

andreeaflorescu:
> Hi all, I've updated the PR, answered the previous comments, and opened some new issues. Would be useful to merge the current iteration, so we can add some more consolidation logic and device simplification on top of these changes. Here are some of the current concerns:
> 
> * The unit tests for the `Tap` struct taken from Firecracker/crosvm require elevated privileges to run, as you have already noticed. Given that we'd like `Tap` to be a separate abstraction (#94), should we just consider the current implementation essentially immutable after we've brought it over and not include those tests as well, or should we temporarily run the tests with more privileges (if they are not so already)?

I think we are more likely to write tests for uncover lines than to come back and re-write the tests such that they do not require sudo (at least this seems to be the case for past experiences). I would prefer us to be in a place where tests can run without polluting the environment.

> * Not sure how creating and tearing down TAP interfaces should exist within the current integration testing framework (as it's required for integration testing). Does anyone have a good starting idea here? Should we create an issue and solve that separately going forward?

We should add an issue for this. Ideally tests will not leave tap interfaces even when the run fails.

aghecenco (approved):
LGTM. The tests pass on CI because they already run privileged there; I think we shouldn't invest more in fixing the privilege problem until #94 is done, and I prefer missing tests to overly complex ones - at least until some of the complexity goes away.

src/devices/src/virtio/net/device.rs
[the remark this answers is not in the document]
lauralt: I'll approve once this comment is addressed. Thanks!

lauralt (approved):

alexandruag:
I removed the `Tap` tests that required special permissions, and I'll add any missing issues for outstanding work items later today. Please have another look, and thank you!

aghecenco (approved):

PR #50, opened 2020-12-16
Rework the vmm & vm-vcpu crates to exit cleanly

andreeaflorescu:
Besides the changes required for implementing clean shut down, there are also a few other fixes:
- added a thread barrier so that we make sure that all vcpu threads start running at the same time
- fix the APIC id mismatch error that was showing when booting; this was caused by a cpuid leaf not being properly set
-  some other tiny fixes from a previous PR


src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -244,81 +245,91 @@ impl KvmVcpu {
...
-                        println!("Guest shutdown: {:?}. Bye!", exit_reason);
-                        if stdin().lock().set_canon_mode().is_err() {
-                            eprintln!("Failed to set canon mode. Stdin will not echo.");
+        loop {
+            match self.vcpu_fd.run() {
+                Ok(exit_reason) => {
+                    match exit_reason {
+                        VcpuExit::Shutdown | VcpuExit::Hlt => {
+                            println!("Guest shutdown: {:?}. Bye!", exit_reason);
+                            if stdin().lock().set_canon_mode().is_err() {
```
aghecenco, line 258: I think this should be moved in the vmm code because
* the vcpu shouldn't care about the terminal's mode, it's an application setting
* it's not even needed if the serial console is absent, or has a different input sink than stdin (not the case for the present reference VMM but should still be configurable, I'll add an issue)

Previously there was no other chance to reset the terminal settings because the process exited abruptly here.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -244,81 +245,91 @@ impl KvmVcpu {
...
                                 .lock()
                                 .unwrap()
-                                .pio_write(PioAddress(addr), data)
+                                .mmio_read(MmioAddress(addr), data)
                                 .is_err()
                             {
-                                eprintln!("Failed to write to serial port");
+                                eprintln!("Failed to read from mmio");
                             }
-                        } else if addr == 0x060 || addr == 0x061 || addr == 0x064 {
```
aghecenco, line 269: I'd keep these around even though they're not called yet; it still exemplifies what to do for i8042 exits, and we'll get to them anyway soon.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -244,81 +245,91 @@ impl KvmVcpu {
...
-                    _ => {
-                        // Unhandled KVM exit.
+                }
+                Err(e) => {
+                    // During boot KVM can exit with `EAGAIN`. In that case, do not
+                    // terminate the run loop.
+                    if e == Errno::new(libc::EAGAIN) {
+                    } else {
+                        eprintln!("Emulation error: {}", e);
+                        break;
```
aghecenco, line 333: nit
```suggestion
                    if e != Errno::new(libc::EAGAIN) {
                        eprintln!("Emulation error: {}", e);
                        break;
                    }
```

src/vm-vcpu/src/vm.rs
```diff
@@ -58,16 +58,29 @@ pub enum Error {
 /// Dedicated [`Result`](https://doc.rust-lang.org/std/result/) type.
 pub type Result<T> = std::result::Result<T, Error>;
 
-impl KvmVm {
+/// Trait that allows the VM to signal that the VCPUs have exited.
+///
+// This trait needs Clone because each VCPU needs to be able to call the `kick` method.
```
aghecenco, line 64: nit
```suggestion
// This trait needs Clone because each VCPU needs to be able to call the `kick` function.
```

src/vm-vcpu/src/vm.rs
```diff
@@ -58,16 +58,29 @@ pub enum Error {
...
+pub trait ExitHandler: Clone {
```
aghecenco, line 65: The trait *itself* doesn't _need_ `Clone` afaict, but instead `Clone` needs to be trait-bounded to the `KvmVm` struct that implements it. So you can leave `ExitHandler` with no dependencies and move the bound here:
```rust
impl<EH: 'static + ExitHandler + Send + Clone> KvmVm<EH> {
```
However, since the exit handler **has** to be `Clone`able anyway because by design it's something the vmm thread and the vcpu threads share, this is more of a nit and I don't feel strongly about where the bound sits either way.

src/vm-vcpu/src/vm.rs
```diff
@@ -194,17 +210,34 @@ mod tests {
     use crate::vm::{Error, KvmVm, VmState};
     use kvm_bindings::CpuId;
     use kvm_ioctls::Kvm;
+    use std::sync::atomic::{AtomicBool, Ordering};
```
aghecenco, line 216: nit: reorder the `use`s.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: I was also thinking that the exit handling should be part of the VMM. What I was planning on doing is a bit different though. I think even what vcpu exists we want to handle is the responsibility of the VMM, and was thinking about moving the full processing there. A try of doing so is in one of my branches: https://github.com/andreeaflorescu/vmm-reference/commit/000b41fe257da6b488c09aef6004ff972ecf2f66

I pushed this change a bit more into the future because the implementation of clean exit & pause resume will infer changes into the exit handler as well.

Would you mind leaving it as is for now, and re-discuss this when I open the other PR that treats all the vcpu exits outside of this crate?

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: They're still there, but GitHub is showing the diff a bit weirdly. They're at line 323 now.

src/vm-vcpu/src/vm.rs
andreeaflorescu: That makes sense. I don't know which is the most idiomatic way (probably your suggestion). The reason I added it as a trait bound is because I wanted to make it clear for the implementers what constraints are needed, and it was also easier to document in the code that clone is needed because the VM is cloning the ExitHandler before sending it to the Vcpus.

andreeaflorescu:
This still does not work for more than 1 vcpu because only one vcpu will exit with shutdown/halt. All the other vcpus need to be kicked. This bug did not surface because we aren't actually joining the vcpu threads, so we didn't see that the vmm-reference process is not exiting.

To get this to be fully functional, kicking the vcpus needs to be implemented, and we also need to have a shared state between vcpus by which we can ensure that when one of them is halted, all the others exit as well. This part is still WIP, and since the functionality is not completely correct, I'll convert this back to an RFC.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: In the next commits this is updated to account for libc::EINTR as well. If that's fine with you, I'll not update it in this commit as well.

andreeaflorescu:
> To get this to be fully functional, kicking the vcpus needs to be implemented, and we also need to have a shared state between vcpus by which we can ensure that when one of them is halted, all the others exit as well. This part is still WIP, and since the functionality is not completely correct, I'll convert this back to an RFC.

:arrow_up: This is now done. The one thing that I could not figure out is a way to kick vcpus such that when one is exiting with shutdown/halt, all the others follow without the need to manually call shutdown. The problem comes when wanting to call kick once one vcpu is exiting because this would need to be called from the vcpu thread.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -61,6 +72,20 @@ pub struct VcpuState {
     pub cpuid: CpuId,
 }
 
+/// Represents the current run state of the VCPUs.
+#[derive(Default)]
+pub struct VcpuRunState {
```
alexandruag, line 77: Not sure what the best state representation is (having a `VmRunState` in `VcpuRunState` feels a bit out of place) or whether a `Condvar` is the best primitive for communication, but we can definitely keep things like this for now and get a better idea as we start working on pause/resume and state save/restore. A couple of things to keep in mind are:

- Once operations such as pause/resume become available, we have to watch out for potential race conditions and state inconsistency (i.e. what happens when pause/unpause are called in random quick succession, given that communication with the vcpus isn't instant or deterministic).
- Data probably has to be passed around when saving state.

A potential alternative is to look at using [channels](https://doc.rust-lang.org/std/sync/mpsc/fn.channel.html) which provide synchronized communication and also allow sending data. This is just something to consider as we move forward.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -240,16 +268,48 @@ impl KvmVcpu {
...
-            self.configure_regs(instruction_pointer)?;
-            self.running = true;
-        }
-        self.run_barrier.wait();
+        self.configure_regs(instruction_pointer)?;
+        self.init_tls()?;
 
+        self.run_barrier.wait();
         loop {
+            let mut interrupted_by_signal = false;
```
alexandruag, line 312: Not really important, but I think renaming `interrupted_by_signal` to just `interrupted` sounds better.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -319,23 +380,63 @@ impl KvmVcpu {
...
+                            // and enter the kvm run loop.
+                            break;
+                        }
+                        VmRunState::Suspending => {
+                            // The VM is suspending. We run this loop until we get a different
+                            // state.
+                        }
+                        VmRunState::Exiting => {
+                            // The VM is exiting. We also exit from this VCPU thread.
+                            return Ok(());
```
alexandruag, line 421: We can use a `break` + `'label` here (example [here](https://doc.rust-lang.org/rust-by-example/flow_control/loop/nested.html)) to exit the outer loop. This is not super useful right now, but I think it's more idiomatic and would help if we want to do more than just return `Ok(())` before leaving the method (otherwise we'd have to duplicate that logic both here and before the `Ok(())` on line 432).

src/vmm/src/lib.rs
```diff
@@ -137,9 +141,49 @@ pub struct VMM {
...
     block_devices: Vec<Arc<Mutex<Block>>>,
 }
 
+// The `VmmExitHandler` is used as the mechanism for exiting from the event manager loop.
+// The Vm is notifying us through the `kick` method when it exited. Once the Vm finished
+// the execution, it is time for the event manager loop to also exit. This way, we can
+// terminate the VMM process cleanly.
+struct VmmExitHandler {
+    exit_event: EventFd,
+    keep_running: AtomicBool,
```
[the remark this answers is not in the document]
andreeaflorescu: I didn't find another way of interrupting the event manager run loop without using an extra field. Do you have a suggestion for how can a write on EventFd would trigger an exit of the loop?

src/vmm/src/lib.rs
```diff
@@ -159,15 +203,25 @@ impl TryFrom<VMMConfig> for VMM {
...
         };
-        let vm = KvmVm::new(&kvm, vm_state, &guest_memory)?;
+
+        let wrapped_exit_handler = WrappedExitHandler(Arc::new(Mutex::new(VmmExitHandler {
+            exit_event: EventFd::new(libc::EFD_NONBLOCK).map_err(Error::ExitEvent)?,
+            keep_running: AtomicBool::new(true),
+        })));
+        let vm = KvmVm::new(&kvm, vm_state, &guest_memory, wrapped_exit_handler.clone())?;
+
+        let mut event_manager = EventManager::<Arc<Mutex<dyn MutEventSubscriber + Send>>>::new()
```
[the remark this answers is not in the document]
andreeaflorescu: It doesn't work because on the following line we are adding the `exit_handler` as a subscriber of `event_manager`, which makes Rust think that the type of subscriber is `event_manager::EventManager<Arc<Mutex<VmmExitHandler>>>`  instead of the expected `dyn MutEventSubscriber`.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -3,7 +3,7 @@
 
 use std::io::{self, stdin};
 use std::result;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Barrier, Mutex};
```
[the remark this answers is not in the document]
andreeaflorescu: It is not really straightforward to reproduce the problem. It might actually have an impact when booting a Linux kernel. I'll have to investigate this more.

alexandruag:
Looks good, would be nice to rebase on top of the latest upstream changes as well.

aghecenco (approved):

alexandruag (approved):

PR #52, opened 2020-12-30
Add vsock device support

alexandruag:
This PR adds support for vsock devices, based on the backend and packet parser implementation currently available in Firecracker (some more details about the Firecracker Unix Domain Socket vsock backend are available [here](https://github.com/firecracker-microvm/firecracker/blob/master/docs/vsock.md)). The code is written and brought over on top of #49. 

One way to verify the vsock device operation is to follow the next steps. First, build the vsock helper program from [here](https://github.com/firecracker-microvm/firecracker/blob/master/tests/host_tools/vsock_helper.c) (ideally as a static binary) and copy the binary to the guest filesystem image. After the guest boots with a vsock device added (i.e. by appending something like ` --vsock cid=5,uds_path=/tmp/vsock_uds0` to the vmm command line), start the listener server within the guest using `./vsock_helper echosrv 5252` (5252 is an arbitrary port number). Then, on the host, we can use something like the following script to verify that can send some data over the UDS vsock backend, and get a reply:

```python
import re
from socket import socket, AF_UNIX, SOCK_STREAM

uds_path = '/tmp/vsock_uds0'
port = 5252

sock = socket(AF_UNIX, SOCK_STREAM)
sock.connect(uds_path)

buf = bytearray("CONNECT {}\n".format(port).encode("utf-8"))
assert sock.send(buf) == len(buf)

ack_buf = sock.recv(32)
assert re.match("^OK [0-9]+\n$", ack_buf.decode('utf-8')) is not None

msg = "vsock is working"
msg_buf = bytes(msg, 'utf-8')

sent = sock.send(msg_buf)
assert sent == len(msg_buf)

recv_buf = sock.recv(sent)
assert len(recv_buf) == len(msg_buf)
assert recv_buf == msg_buf

print(recv_buf.decode('utf-8'))
```

We're going to bring over parts of the vsock test infrastructure as well after advancing the discussion around dirty page tracking in `vm-memory` and seeing how the ongoing `vm-virtio` discussions conclude. The former is important because we want to start using `VolatileSlice`s for the `VsockPacket` abstractions, while the latter influence the boiler plate code we have to adapt as part of the testing infrastructure.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -130,9 +130,9 @@ dependencies = [
 
 [[package]]
 name = "log"
-version = "0.4.6"
+version = "0.4.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c84ec4b527950aa83a329754b01dbe3f58361d1c5efacd1f6d68c494d08a17c6"
+checksum = "4fabed175da42fed1fa0746b0ea71f412aa9d35e76e95e59b192c64b9dc2bf8b"
 dependencies = [
  "cfg-if",
 ]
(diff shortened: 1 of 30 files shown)


andreeaflorescu:
Closing this as we are going to add support once we upstream the components in vm-virtio.

Issue #27, opened 2020-11-10
Integration test that verifies the available CPUs from inside the guest


aghecenco:
Add an integration test that boots a guest and verifies that the CPUs by the VMM are indeed available (parsing `/proc/cpuinfo` is a good start).

dodsonmg:
Is this still open?  If so, I'd like to have a go, but might need some help.

- Are you looking for a test to land in `vmm-reference/src/vmm/tests/integration_tests.rs`?
- Are there any similar tests anywhere in `rust-vmm` that parse and pull information from inside the vm for use externally in a test?

andreeaflorescu:
> Is this still open? If so, I'd like to have a go, but might need some help.
> 

Yap, we can help out here.

> * Are you looking for a test to land in `vmm-reference/src/vmm/tests/integration_tests.rs`?

We were looking for a Python integration test. These are placed under [vmm-reference/tests/](https://github.com/rust-vmm/vmm-reference/tree/master/tests). I think it should be possible to do it in Rust as well, but there are a few examples of how to communicate with the VM in Python already if you'd like to reuse those.

> * Are there any similar tests anywhere in `rust-vmm` that parse and pull information from inside the vm for use externally in a test?

The 2 things that we need to do to be able to check the number of cpus are:
- send a command to the VM; an example of this is in the the test that [shuts down the VM](https://github.com/rust-vmm/vmm-reference/blob/1c5ed93058ba83b87a18052a8ef2975752932edd/tests/test_run_reference_vmm.py#L91)
- check the response of the command; we do not have an example here, but I think that as a first iteration we can work with something similar to the [`expect_string`](https://github.com/rust-vmm/vmm-reference/blob/1c5ed93058ba83b87a18052a8ef2975752932edd/tests/test_run_reference_vmm.py#L99) function to read the stdout of the process.

dodsonmg:
Closed by #53?

Issue #26, opened 2020-11-10
Integration test that verifies the free guest memory from inside the guest


aghecenco:
Add an integration test that boots a guest and verifies that the memory configured by the VMM is indeed available (parsing `/proc/meminfo` is a good start).

dodsonmg:
Closed by #53?

PR #53, opened 2021-01-05
Integration tests to check number of vcpus and available memory from within the guest

dodsonmg:
Addresses issues #26 and #27.

### DIFF
--- a/tests/test_run_reference_vmm.py
+++ b/tests/test_run_reference_vmm.py
@@ -37,11 +37,10 @@ def process_exists(pid):
 The following methods would be nice to have a part of class revolving
 around the vmm process. Let's figure out how to proceed here as part
 of the discussion around making the CI/testing easier to use, extend,
-and run locally. 
+and run locally.
 """
 
-
-def start_vmm_process(kernel_path, disk_path):
+def start_vmm_process(kernel_path, disk_path, vcpus = 1):
     # Memory config
     mem_size_mib = 1024
 
@@ -51,7 +50,7 @@ def start_vmm_process(kernel_path, disk_path):
     himem_start = 1048576
 
     # vCPU config
-    num_vcpus = 1
+    num_vcpus = vcpus
 
     build_cmd = "cargo build --release"
     subprocess.run(build_cmd, shell=True, check=True)
--- a/tests/test_run_reference_vmm.py
+++ b/tests/test_run_reference_vmm.py
@@ -98,6 +98,9 @@ def expect_string(vmm_process, expected_string):
         nextline = vmm_process.stdout.readline()
         if expected_string in nextline.decode():
             break
+    # If the expected string is identified and the loop breaks, return the
+    # entire line from stdout
+    return nextline.decode()
 
 def expect_vcpus(vmm_process, expected_vcpus):
     # /proc/cpuinfo displays info about each vCPU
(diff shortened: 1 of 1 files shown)


tests/test_run_reference_vmm.py
```diff
@@ -156,3 +180,22 @@ def test_reference_vmm_num_vcpus(kernel):
         expect_vcpus(vmm_process, expected_vcpus)
 
         shutdown(vmm_process)
+
+@pytest.mark.parametrize("kernel", KERNELS_INITRAMFS)
+def test_reference_vmm_mem(kernel):
+    """Start the reference VMM and verify the amount of memory."""
+
+    mem_sizes_mib = [512, 1024, 2048]
```
andreeaflorescu, line 188: Can you check with a value that is higher that the start of the MMIO Gap address? The value is defined here: https://github.com/rust-vmm/vmm-reference/blob/1c5ed93058ba83b87a18052a8ef2975752932edd/src/vmm/src/lib.rs#L60)

We are creating 2 memory regions when the memory does not fit before the MMIO gap, and we should probably check that we still get the expected amount of memory in that situation.

tests/test_run_reference_vmm.py
```diff
@@ -102,6 +102,21 @@ def expect_string(vmm_process, expected_string):
         if expected_string in nextline.decode():
             break
 
+def expect_vcpus(vmm_process, expected_vcpus):
+    # /proc/cpuinfo displays info about each vCPU
+    vmm_process.stdin.write(b'cat /proc/cpuinfo\n')
+    vmm_process.stdin.flush()
+
+    # The sibling field for each vCPU displays the total number of vCPUs
+    expected_siblings = "siblings:{}".format(expected_vcpus)
```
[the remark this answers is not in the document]
dodsonmg: Yes.  That makes sense.  Will fixup.

tests/test_run_reference_vmm.py
dodsonmg: If I'm reading it right, I need a value > 4GB?

That makes sense.  I can also explicitly test the edge cases.  Will fixup.

andreeaflorescu: Yap, that's correct. We can remove some other values from the `mem_sizes_mib` list so that the test does not take too long.

dodsonmg: Ok.  I'm now testing well below, well above, and then right at the start of the MMIO gap to verify that the partitioning logic in `create_guest_memory()` is working:

https://github.com/rust-vmm/vmm-reference/blob/1c5ed93058ba83b87a18052a8ef2975752932edd/src/vmm/src/lib.rs#L222-L230

If that's overkill, I can pull some of it out.

andreeaflorescu: I think that's fine for now. In the future we can have some more optimization so the tests take less (like for example grouping together the tests for memory size, vcpu count, and maybe disk?), but for now it's not necessary as we're also still figuring out a testing framework.

andreeaflorescu (approved):

tests/test_run_reference_vmm.py
```diff
@@ -102,6 +96,52 @@ def expect_string(vmm_process, expected_string):
...
+            nextline = "".join(nextline.split())
+            # extract the number of vCPUs
+            actual_vcpus = int(nextline.split(':')[1])
+            break
+
+    assert actual_vcpus == expected_vcpus, \
+            "Expected {}, found {} vCPUs".format(expected_vcpus, actual_vcpus)
+
+def expect_mem(vmm_process, expected_mem_mib):
+    expected_mem_kib = expected_mem_mib * 1000
```
lauralt, line 120: Shouldn't 1000 be 1024?

tests/test_run_reference_vmm.py
```diff
@@ -102,6 +96,52 @@ def expect_string(vmm_process, expected_string):
...
+        nextline = vmm_process.stdout.readline().decode()
+        if "MemTotal" in nextline:
+            break
+
+    # Remove whitespace and extract the memory value
+    # Expected format is 'MemTotal: [value] kB'
+    mem_total = nextline.split()[1]
+
+    # Expect the difference between the expected and actual memory
+    # to be less than 10%
```
andreeaflorescu, line 140: On a second thought, 10% seems like a lot. Would a fixed diff in KB work as well?

dodsonmg:
I hadn't noticed the recent commits to `master`, so I just pulled them in, rebased, and force-pushed, but the CI failed almost immediately.

I don't seem to be able to see the test results (other than a big, red circle-X).  The tests are passing for me locally, though, so if someone could give me a hint, I would be grateful.

andreeaflorescu:
> I hadn't noticed the recent commits to `master`, so I just pulled them in, rebased, and force-pushed, but the CI failed almost immediately.
> 
> I don't seem to be able to see the test results (other than a big, red circle-X). The tests are passing for me locally, though, so if someone could give me a hint, I would be grateful.

@dodsonmg we had some issues with the CI, but they should be fixed now. The tests in this PR are still failing, but it seems that it is a change related to this PR. Can you remove the rust-vmm-ci update from your first commit? I have a hunch that might be the problem.

lauralt:
Besides Andreea's comment, changes look good to me :+1:.

lauralt (approved):

dodsonmg:
Whew... Sorry that took so long.  Thanks for helping me through!

andreeaflorescu (approved):

PR #56, opened 2021-01-12
Issue 55

### LINKED ISSUE #55: Implement a more generic parser for command line arguments
We currently duplicate the parsing logic in the `try_from` functions of `MemoryConfig`, `VcpuConfig`, `KernelConfig`, `NetConfig`, `BlockConfig`. As we add more devices and configuration, this duplication generates too much boilerplate code.

We should try to see if we can somehow reduce this duplication (without adding new rust crates dependencies).

FedericoPonzi:
Fixes #55 

This is still a draft PR. While most of the code is there, I would like to add some more testing for the code I've added.
    
This refactoring aims to deduplicate the parsing effort among the config's
structs and ease the addition of new parameters and structs.

The ArgParsing trait is used to offer a parse method, which needs to be
implemented for each type one wants to parse. If there are special parsing
cases (e.g. u8 > 0) it's also possible to introduce this by creating a new
wrapper struct and implement the ArgParsing accordingly.

For avoiding code duplication, `argparsing` macro is introduced to avoid
copy pasting identical implementation for numberic types.

Finally, the newly introduced ParsingError enum allows for more precise
error reporting, and more on point error reporting (please check the updated
tests cases with example of possible errors).

### DIFF
--- a/coverage_config_x86_64.json
+++ b/coverage_config_x86_64.json
@@ -1,5 +1,5 @@
 {  
-  "coverage_score": 79.7,
+  "coverage_score": 79.8,
   "exclude_path": "msr_index.rs,mpspec.rs,tests/",
   "crate_features": ""
 }
--- a/src/api/src/lib.rs
+++ b/src/api/src/lib.rs
@@ -157,7 +157,7 @@ mod tests {
         ])
         .is_err());
 
-        // Invalid memory config: missing value for `size_mib`.
+        // Memory config: missing value for `size_mib`, use the default
         assert!(CLI::launch(vec![
             "foobar",
             "--memory",
@@ -167,7 +167,7 @@ mod tests {
             "--kernel",
             "path=/foo/bar,cmdline=\"foo=bar\",himem_start=42",
         ])
-        .is_err());
+        .is_ok());
 
         // Invalid memory config: unexpected parameter `foobar`.
         assert!(CLI::launch(vec![
@@ -194,7 +194,7 @@ mod tests {
         ])
         .is_err());
 
-        // Invalid kernel config: missing value for `himem_start`.
+        // Kernel config: missing value for `himem_start`, use default
         assert!(CLI::launch(vec![
             "foobar",
             "--memory",
@@ -204,7 +204,7 @@ mod tests {
             "--kernel",
             "path=/foo/bar,cmdline=\"foo=bar\",himem_start=",
         ])
-        .is_err());
+        .is_ok());
 
         // Invalid kernel config: unexpected parameter `foobar`.
         assert!(CLI::launch(vec![
@@ -230,7 +230,7 @@ mod tests {
         ])
         .is_err());
 
-        // Invalid vCPU config: missing value for `num_vcpus`.
+        // vCPU config: missing value for `num_vcpus`, use default
         assert!(CLI::launch(vec![
             "foobar",
             "--memory",
@@ -240,7 +240,7 @@ mod tests {
             "--kernel",
             "path=/foo/bar,cmdline=\"foo=bar\",himem_start=42",
         ])
-        .is_err());
+        .is_ok());
 
         // Invalid vCPU config: unexpected parameter `foobar`.
         assert!(CLI::launch(vec![
--- a/src/vmm/src/config/arg_parser.rs
+++ b/src/vmm/src/config/arg_parser.rs
@@ -1,0 +1,129 @@
+use crate::config::arg_parser::CfgArgParseError::UnknownArg;
+use std::collections::HashMap;
+use std::fmt;
+use std::fmt::Formatter;
+use std::str::FromStr;
+
+#[derive(Debug, PartialEq)]
+pub(super) enum CfgArgParseError {
+    /// Parsing failed, param and error.
+    ParsingFailed(&'static str, String),
+    UnknownArg(String),
+}
+
+impl fmt::Display for CfgArgParseError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::ParsingFailed(param, err) => {
+                write!(f, "Param '{}', parsing failed: {}", param, err)
+            }
+            Self::UnknownArg(err) => write!(f, "Unknown arguments found: '{}'", err),
+        }
+    }
+}
+
+pub(super) struct CfgArgParser {
+    args: HashMap<String, String>,
+}
+
+impl CfgArgParser {
+    pub(super) fn new(input: String) -> Self {
+        let args = input
+            .split(',')
+            .filter(|tok| !tok.is_empty())
+            .map(|tok| {
+                let mut iter = tok.splitn(2, '=');
+                let param_name = iter.next().unwrap();
+                let value = iter.next().unwrap_or("").to_string();
+                (param_name.to_lowercase(), value)
+            })
+            .collect();
+        Self { args }
+    }
+
+    /// Retrieves the value of `param`, consuming it from `Self`.
+    pub(super) fn value_of<T: FromStr>(
+        &mut self,
+        param_name: &'static str,
+    ) -> Result<Option<T>, CfgArgParseError>
+    where
+        <T as FromStr>::Err: std::fmt::Display,
+    {
+        match self.args.remove(param_name) {
+            Some(value) if !value.is_empty() => value
+                .parse::<T>()
+                .map_err(|err| CfgArgParseError::ParsingFailed(param_name, err.to_string()))
+                .map(Some),
+            _ => Ok(None),
+        }
+    }
+
+    /// Checks if all params were consumed.
+    pub(super) fn all_consumed(&self) -> Result<(), CfgArgParseError> {
+        if self.args.is_empty() {
+            Ok(())
+        } else {
+            Err(UnknownArg(self.to_string()))
+        }
+    }
+}
+
+impl fmt::Display for CfgArgParser {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            self.args
+                .keys()
+                .map(|val| val.as_str())
+                .collect::<Vec<&str>>()
+                .join(", ")
+        )
+    }
+}
+#[cfg(test)]
+mod tests {
+    use crate::config::arg_parser::{CfgArgParseError, CfgArgParser};
+    use std::num::NonZeroU8;
+    use std::path::PathBuf;
+
+    #[test]
+    fn test_cfg_arg_parse() -> Result<(), CfgArgParseError> {
+        let input_params = "path=/path,string=HelloWorld,int=123,u8=1".to_string();
+        let mut arg_parser = CfgArgParser::new(input_params);
+
+        // No parameter was consumed yet
+        assert!(arg_parser.all_consumed().is_err());
+
+        assert_eq!(
+            arg_parser.value_of::<PathBuf>("path")?.unwrap(),
+            PathBuf::from("/path")
+        );
+        assert_eq!(
+            arg_parser.value_of::<String>("string")?.unwrap(),
+            "HelloWorld".to_string()
+        );
+        assert_eq!(
+            arg_parser.value_of::<NonZeroU8>("u8")?.unwrap(),
+            NonZeroU8::new(1).unwrap()
+        );
+        assert_eq!(arg_parser.value_of::<u64>("int")?.unwrap(), 123);
+
+        // Params now is empty, use the Default instead.
+        let default = 12;
+        assert_eq!(arg_parser.value_of("int")?.unwrap_or(default), default);
+
+        // Params is empty and no Default provided:
+        assert!(arg_parser.value_of::<u64>("int")?.is_none());
+
+        // All params were consumed:
+        assert!(arg_parser.all_consumed().is_ok());
+
+        let input_params = "path=".to_string();
+        assert!(CfgArgParser::new(input_params)
+            .value_of::<String>("path")?
+            .is_none());
+
+        Ok(())
+    }
+}
--- a/src/vmm/src/config/mod.rs
+++ b/src/vmm/src/config/mod.rs
@@ -3,10 +3,15 @@
 
 use std::convert::TryFrom;
 use std::fmt;
+use std::num;
 use std::path::PathBuf;
 use std::result;
 
+use arg_parser::CfgArgParser;
+
 use super::{DEFAULT_HIGH_RAM_START, DEFAULT_KERNEL_CMDLINE};
+
+mod arg_parser;
 
 /// Errors encountered converting the `*Config` objects.
 #[derive(Debug, PartialEq)]
@@ -21,6 +26,24 @@ pub enum ConversionError {
     ParseNetwork(String),
     /// Failed to parse the string representation for the block.
     ParseBlock(String),
+}
+
+impl ConversionError {
+    fn new_kernel<T: fmt::Display>(err: T) -> Self {
+        Self::ParseKernel(err.to_string())
+    }
+    fn new_memory<T: fmt::Display>(err: T) -> Self {
+        Self::ParseMemory(err.to_string())
+    }
+    fn new_vcpus<T: fmt::Display>(err: T) -> Self {
+        Self::ParseVcpus(err.to_string())
+    }
+    fn new_block<T: fmt::Display>(err: T) -> Self {
+        Self::ParseBlock(err.to_string())
+    }
+    fn new_network<T: fmt::Display>(err: T) -> Self {
+        Self::ParseNetwork(err.to_string())
+    }
 }
 
 impl fmt::Display for ConversionError {
@@ -48,23 +71,16 @@ impl TryFrom<String> for MemoryConfig {
 
     fn try_from(mem_cfg_str: String) -> result::Result<Self, Self::Error> {
         // Supported options: `size=<u32>`
-        let mem_cfg_str_lower = mem_cfg_str.to_lowercase();
-        let tokens: Vec<&str> = mem_cfg_str_lower
-            .split('=')
-            .filter(|tok| !tok.is_empty())
-            .collect();
-        if tokens.len() != 2 {
-            return Err(ConversionError::ParseMemory(mem_cfg_str));
-        }
-        if tokens[0] != "size_mib" {
-            return Err(ConversionError::ParseMemory(tokens[0].to_string()));
-        }
-        tokens[1]
-            .parse::<u32>()
-            .map(|mem_size_mib| MemoryConfig {
-                size_mib: mem_size_mib,
-            })
-            .map_err(|_| ConversionError::ParseMemory(tokens[1].to_string()))
+        let mut arg_parser = CfgArgParser::new(mem_cfg_str);
+
+        let size_mib = arg_parser
+            .value_of("size_mib")
+            .map_err(ConversionError::new_memory)?
+            .unwrap_or(256);
+        arg_parser
+            .all_consumed()
+            .map_err(ConversionError::new_memory)?;
+        Ok(MemoryConfig { size_mib })
     }
 }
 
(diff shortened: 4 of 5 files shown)


FedericoPonzi:
Unfortunately I cannot access buldkite, I'm not sure what's currently failing.

andreeaflorescu:
> Unfortunately I cannot access buldkite, I'm not sure what's currently failing.

Sorry about that, it was a misconfiguration on our part. You should now be able to see the results of the pipeline if you click on the link.

FedericoPonzi:
It works now, thanks!

src/vmm/src/config.rs
```diff
@@ -23,6 +94,42 @@ pub enum ConversionError {
...
+impl ToString for ParsingError {
+    fn to_string(&self) -> String {
+        match self {
+            Self::MissingRequired(err) => format!("Missing required argument: {}", err),
+            Self::ParsingFailed(param, err) => {
+                format!("Param '{}', parsing failed: {}", param, err)
+            }
+        }
+    }
+}
```
gabhijit, line 114: As per [this recommendation](https://doc.rust-lang.org/std/string/trait.ToString.html) will it not be better to `impl Display` for this ?

src/vmm/src/config.rs
```diff
@@ -1,13 +1,84 @@
...
+        fn parse(raw: String) -> Result<Self, ParseIntError> {
+            raw.parse()
+        }
+    })*)
+}
+
+argparsing! { u32 u64 NonZeroU8 }
+
+impl ArgParsing for PathBuf {
+    fn parse(raw: String) -> Result<Self, ParseIntError> {
```
gabhijit, line 30: May be define a `ParseError` `Enum` and have different variants?

src/vmm/src/config.rs
```diff
@@ -1,13 +1,84 @@
...
+}
+impl ArgParsing for String {
+    fn parse(raw: String) -> Result<Self, ParseIntError> {
+        Ok(raw)
+    }
+}
+
+fn tokenize(input: String) -> HashMap<String, String> {
+    input
+        .split(",")
```
gabhijit, line 42: nit: clippy warning - use a `','`

src/vmm/src/config.rs
```diff
@@ -1,13 +1,84 @@
...
+        .filter(|tok| !tok.is_empty())
+        .map(|tok| {
+            let mut iter = tok.split("=");
```
gabhijit, line 45: nit: clippy warning - use a '=' [explaination](https://rust-lang.github.io/rust-clippy/master/index.html#single_char_pattern)

src/vmm/src/config.rs
gabhijit: Ignore above comment, I see - that's a `trait` not a `struct`, but still the error type be a different Error. May be an associated Error type that is an `enum` and that uses different variants?

gabhijit (commented):
Also, if you can add a general high level approach to how you are doing this, it would help. Usually the detailed paragraph  of your `git commit` message should do this and get added as a detailed description of the Pull Request. This way it helps to understand what you are doing.

src/vmm/src/config.rs
FedericoPonzi: I already have added a `ParsingError` enum with different variants. 

[Here](https://github.com/rust-vmm/vmm-reference/pull/56/files#diff-919306a64a3d1ead59ffc1193399bb62123b222fc3a748b5d9dc3a34d6bdd6bfR60) I'm taking the possible error from the parse function, and wrap it in a ParsingError variant which also holds the parameter and the error. To do what you suggested, I would need to create yet another enum variant only for the parse function. I decide to use the `num::ParseIntError` as error type instead, because numeric conversion are the only parsing which can fail (parsing String -> String and String -> PathBuf is infallible).

If you think it's better, I can use the ParsingError variant, and pass the param_name as input to the parse function. In this way, each `parse` implementation can use it for building an helpful error. WDYT?

gabhijit: Yes, I still can't think of what is a 'right' way, but something that returned a `String` as a result, returning `ParseIntError` didn't seem too right for me. So this `ParsingError` should be fine for now.

src/vmm/src/config.rs
```diff
@@ -1,13 +1,100 @@
...
+use std::collections::HashMap;
 use std::convert::TryFrom;
 use std::fmt;
+use std::num;
 use std::path::PathBuf;
 use std::result;
 
 use super::{DEFAULT_HIGH_RAM_START, DEFAULT_KERNEL_CMDLINE};
 
+trait ArgParsing {
```
gabhijit, line 13: Do you think  it's a good idea to use a name like `CfgArgParser` and `CfgArgParseError` ? They appear specific to config parsing, so slightly more 'explicit' sounds better.

gabhijit: And then also may be rename the `fn` to something like `cfg_parse` as this is really doing that and `parse` is a standard library function  for [`str` ish types](https://doc.rust-lang.org/std/string/struct.String.html#method.parse) ?

src/vmm/src/config.rs
```diff
@@ -1,13 +1,100 @@
...
+fn tokenize(input: String) -> HashMap<String, String> {
+    input
+        .split(',')
+        .filter(|tok| !tok.is_empty())
+        .map(|tok| {
+            let mut iter = tok.split('=');
+            let param_name = iter.next().unwrap();
+            let value = iter.collect::<Vec<&str>>().join("=");
+            (param_name.to_lowercase(), value)
+        })
```
gabhijit, line 49: I believe this logic is to be able to handle something like `cmdlin=\"foo=bar\"`? Is that right? How about something like the following - because this code is a bit difficult to read. 

```rust 

.map(|tok| {
            let mut iter = tok.splitn(2, '=');
            let param_name = iter.next().unwrap();
            let value = iter.next().unwrap().to_string();
            (param_name.to_lowercase(), value)
        })
```

I quickly tried this out in [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4c5c8b81a792954edfec9c45aa33eca8)  and seems to work. If this is what you are expecting.

src/vmm/src/config.rs
```diff
@@ -1,13 +1,100 @@
...
+        Ok(PathBuf::from(raw))
+    }
+}
+impl ArgParsing for String {
+    fn parse(_param_name: &'static str, raw: String) -> Result<Self, ParsingError> {
+        Ok(raw)
+    }
+}
+
+fn tokenize(input: String) -> HashMap<String, String> {
```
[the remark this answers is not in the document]
gabhijit: Also, when I think a bit more about it, why not fail early here? `eg.` if you find something like `path=` or `path`, return an error here. Or may be not. Because adding that logic here would complicate handling values without `=` if they ever will be required later on. So perhaps the way it is is fine.

src/vmm/src/config.rs
```diff
@@ -79,26 +172,12 @@ impl TryFrom<String> for VcpuConfig {
...
-            .map_err(|_| ConversionError::ParseVcpus(tokens[1].to_string()))?;
-        if vcpu_config.num == 0 {
-            return Err(ConversionError::ParseVcpus(tokens[1].to_string()));
-        }
-        Ok(vcpu_config)
+        // Supported options: `num=<u8>`
+        let mut tokens = tokenize(vcpu_cfg_str);
+        let num = parse_arg(&mut tokens, "num", Some(num::NonZeroU8::new(1).unwrap()))
+            .map_err(ConversionError::new_vcpus)?;
+        check_all_consumed(&tokens).map_err(ConversionError::new_vcpus)?;
```
gabhijit, line 179: I think one can get rid of calling this function completely. 

At the end of all the calls to `parse_arg` calls. Just check the `tokens` is empty and if not create an error and return. That's perhaps still some boiler plate, but won''t require to make this function call.

src/vmm/src/config.rs
```diff
@@ -48,23 +153,11 @@ impl TryFrom<String> for MemoryConfig {
...
-        if tokens[0] != "size_mib" {
-            return Err(ConversionError::ParseMemory(tokens[0].to_string()));
-        }
-        tokens[1]
-            .parse::<u32>()
-            .map(|mem_size_mib| MemoryConfig {
-                size_mib: mem_size_mib,
-            })
-            .map_err(|_| ConversionError::ParseMemory(tokens[1].to_string()))
+        let mut tokens = tokenize(mem_cfg_str);
```
gabhijit, line 156: Related to `tokenize` comment above, how about calling it a - `config_map` or even more explicit like `mem_cfg_map`, `net_cfg_map` etc. ?

gabhijit (commented):
Mostly 'nits' related to function naming etc. Couple of suggestions.

src/vmm/src/config.rs
gabhijit: I am also thinking, if it makes sense to not have this all heavy lifting of defining a trait etc. A simple generic function should be able to take care of this. Something like the following - 

```rust 

fn cfg_param_parse<T: FromStr>(name: &str, value: String)  -> Result<T, <T as FromStr>::Err> {
    value.parse::<T>()
}
```
And then just `map_err` returned error if any to `ConversionError`. I believe all the values that will be parsed will be `FromStr` (Most of them will simply be strings, leaving behind some `int` and `PathBuf` types.) This should make the whole implementation considerably simple. 

in this [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f84583eb6866afca0b4e73233972735a)

FedericoPonzi: > I am also thinking, if it makes sense to not have this all heavy lifting of defining a trait etc. A simple generic function should be able to take care of this. Something like the following -
> 
> ```rust
> fn cfg_param_parse<T: FromStr>(name: &str, value: String)  -> Result<T, <T as FromStr>::Err> {
>     value.parse::<T>()
> }
> ```
> 
> And then just `map_err` returned error if any to `ConversionError`. I believe all the values that will be parsed will be `FromStr` (Most of them will simply be strings, leaving behind some `int` and `PathBuf` types.) This should make the whole implementation considerably simple.
> 
> in this [playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=f84583eb6866afca0b4e73233972735a)

This is actually a very good idea, thanks! Even if the current proposal should be more easily extensible for types that might need to do something more complex other then `parse`, I currently don't see any reason not to use your proposal instead. Thanks!

src/vmm/src/config.rs
FedericoPonzi: I'm not sure if I got what you mean, this function call is exactly doing that. I think it's cleaner to have the function call for this check and create the related error there (once for every TryFrom functions). If we ever need to update the error format, one would just need to update it in that function, once. Did I address correctly your comment?

src/vmm/src/config.rs
FedericoPonzi: I see your point on the map vs vec, but "config_map" is a bit misleading IMO. This map contains tokens that if correctly consumed will lead to a config_map, but at this point that map contains just an intermediate (possibly invalid) representation. Would `token_map` be suitable for you? Also, for reducing cognitive load and ease copy/pasting, I would call it in a "generic" way, thus "config_map" or "tokens_map" and so on.

src/vmm/src/config.rs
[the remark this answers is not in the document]
FedericoPonzi: That's exactly what I had in mind. This function should just tokenize the input string by `=`, and this operation should always succeed IMO. If a required parameter is missing the conversion will eventually fail anyway.

Also, as you correctly mentioned, not failing early allows to quickly add support for flags, e.g. `--path` (if they will ever be needed).

Finally, I've renamed the function to `cfg_args_to_map`. I don't see any problem with a tokenize function returning an HashMap TBH, but I don't have a strong opinion on this. Thanks for all your comments so far!

src/vmm/src/config.rs
gabhijit: well - I don't think a generic name is needed, since this call is in the `try_from` for each of the configs - so something like `mecfg_params_map` or better still `memcfg_params` is good enough because the original function already tells it's a map.  I would've called it something like `memcfg_params = cfg_args_to_map(...)`  - clearly conveys. But it's just an opinion. I am fine, if you prefer `token_map`.

src/vmm/src/config.rs
gabhijit: Yes, you are right but something that can be done with a simple `if` / `else`. But I am not overly opinionated about it, the function call is just fine too. In fact - in general - unprocessed arguments should ideally be simply `log::warn` and need not be a failure as they are not doing anything wrong or unpleasantly surprising (and this is why I simply prefer `if`/`else` - in fact only `if`). But the way it is - is just fine too.

andreeaflorescu (commented):
I have a few high level suggestions before digging into the details:
- Can we move the argument parsing to another file? The `config.rs` file is pretty bloated by now, and I think it should only have the configuration structure defines, and the implementation of `try_from`. Since the argument parsing is now going into a generic direction, we could move it to its own module.
- I would recommend defining a `ArgumentParser` (or some better name) structure that creates the `HashMap` with parameters. With this approach we achieve a few things:
  - functions are contextualized (we no longer have functions that are defined directly in the module - phantom functions)
  - we do not need to pass HashMap to all function defintions
 

I was thinking about a structure like so:

```rust
struct ArgParser {
    args: HashMap<String, String>,
}

impl ArgParser {
  fn new(input: String) -> Self {
      // this will be the same as `cfg_args_to_map`
  }
    /// Retrieves the value of `param`, consuming it from `Self`.
    fn value_of<T: FromStr>(&mut self, param: &'static str) -> Result<Option<T>, CfgArgParseError>
    where
        <T as FromStr>::Err: std::fmt::Display,
    {
         // This will be the same as `cfg_param_parse`.
         .....
    }

    fn all_consumed(&self) -> Result<(), String> {
    }
}
````

What do you think?

File: src/vmm/src/config.rs [hunk 10]
```diff
@@ -1,13 +1,87 @@
...
+/// This function is currently a thing wrapper around the FromStr function.
+/// # Arguments
+///
+/// * `params` - An hashmap resulting from `cfg_param_parse` call.
+/// * `param_name` - The name of the parameter to parse.
+/// * `default` - A default value to use if this parameter was not supplied.
+fn cfg_param_parse<T: FromStr>(
+    params: &mut HashMap<String, String>,
+    param_name: &'static str,
+    default: Option<T>,
```
andreeaflorescu, line 38: I believe the default parameter is better handled outside of this function, because this function just parses parameters, and retrieves the value if it exists. The defaults are dependent on the caller, and should be handled in the calling function.

The function would need to return a `Result<Option<T>, CfgArgParseError>`, and the caller would need to provide a default as follows:
```rust
cfg_param_parse(
      tokens,
      "mem_size_mib",
) .map_err(ConversionError::new_memory)?.unwrap_or(256);
```

src/vmm/src/config.rs
```diff
@@ -1,13 +1,87 @@
...
+fn check_all_consumed(params: &HashMap<String, String>) -> Result<(), String> {
+    if params.is_empty() {
+        Ok(())
+    } else {
+        let arguments = params
+            .keys()
+            .map(|val| val.as_str())
+            .collect::<Vec<&str>>()
+            .join(", ");
+        Err(format!("Uknown arguments found: '{}'", arguments))
```
andreeaflorescu, line 62: nit: spelling error in "Uknown".

src/vmm/src/config.rs
```diff
@@ -1,13 +1,87 @@
...
+    }
+}
+
+#[derive(Debug, PartialEq)]
+enum CfgArgParseError {
```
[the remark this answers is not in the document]
gabhijit: +1 . This is a lot more cleaner. So that next person does not have to worry about the derails.

src/vmm/src/config.rs
gabhijit: Yes, this is also way better. I was also thinking - perhaps there's a more readable way to do this, but couldn't think of one. This is neat!

lauralt:
@FedericoPonzi there seems to be a problem with the CI (our guess is that it might be related to the fact that it's a draft PR). The buildkite test result from this page shows an older build... this would be the last build: https://buildkite.com/rust-vmm/vmm-reference-ci/builds/261. Sorry, we're looking into this.

src/vmm/src/config.rs
FedericoPonzi: Thanks for this suggestion(s)! I initially thought about it in this way to further reduce code duplication. Your proposal works well for arguments which have the default parameter (like this one). For args which don't have a default it would need to build a new Error with a "missing parameter" error. 

This is an example for `path` argument of Kernel config:
```
        let path = arg_parser
            .value_of("path")
            .map_err(ConversionError::new_kernel)?
            .ok_or_else(|| ConversionError::new_memory("Missing parameter: path"))?;
```
When querying a missing value from the hashmap, the return value is "Optional" only if we have a default value for it. Otherwise, it should be a Result because missing from the hashmap is an Error. 

* If the additional overhead is ok, I can proceed with that. 
* Otherwise, what if argparse provides another function similar which returns an Error if the parameter is missing, thus with type `Result<T, CfgParseError>`? It can also live as a function outside the ArgParser struct if we don't want it there. I would call it something like `required_value_of()`.

src/vmm/src/config.rs
FedericoPonzi: It was just my impression that the parsing should fail if the tokens are not all consumed - I was thinking about mistypings - but I think you're right, maybe a `log::warn` could be enough. @andreeaflorescu can you provide some input on this?

src/vmm/src/config.rs
andreeaflorescu: > When querying a missing value from the hashmap, the return value is "Optional" only if we have a default value for it. Otherwise, it should be a Result because missing from the hashmap is an Error.
> 

I believe it's the caller responsibility to decide weather a `None` is an `Error`. We basically only return errors when there was a value to begin with. From a separation of concerns point of view, I believe it is better to have the occasional duplication if needed.

Another option would be to have a function that treats this scenario in ArgParser. I was thinking of something that just does this for String, as follows:
```rust
    fn value_of_required<T: FromStr>(&mut self, param: &'static str) -> Result<String, CfgArgParseError> {
        self.value_of::<String>(param)?.ok_or(Err(CfgArgParseError::MissingRequired(param.to_string()))
    }
```

And then when parsing the kernel we would have:
```rust
let path = tokens.value_of_required("path")?;
if path.is_empty() {
    .....
}
```
There would still be a bit of duplication, but I would find it acceptable either way. I believe separating concerns is more important than a few duplicated lines.

FedericoPonzi:
@lauralt Thanks for the heads-up, I think you're right and now that the PR is marked ready to review, it is showing the latest build.

src/vmm/src/config.rs
```diff
@@ -48,23 +68,16 @@ impl TryFrom<String> for MemoryConfig {
...
-        if tokens[0] != "size_mib" {
-            return Err(ConversionError::ParseMemory(tokens[0].to_string()));
-        }
-        tokens[1]
-            .parse::<u32>()
-            .map(|mem_size_mib| MemoryConfig {
-                size_mib: mem_size_mib,
-            })
-            .map_err(|_| ConversionError::ParseMemory(tokens[1].to_string()))
+        let mut arg_parser = CfgArgParser::new(mem_cfg_str);
```
gabhijit, line 71: This looks a lot cleaner now! The new struct and associated functions are quite clean right?

src/vmm/src/arg_parser.rs
```diff
@@ -0,0 +1,129 @@
+use crate::arg_parser::CfgArgParseError::UnknownArg;
+use std::collections::HashMap;
+use std::fmt;
+use std::fmt::Formatter;
+use std::str::FromStr;
+
+#[derive(Debug, PartialEq)]
+pub(crate) enum CfgArgParseError {
```
gabhijit, line 8: Wondering if `pub(crate)` is needed. This is very specific to `config` parsing so may be it should be something like `config/mod.rs`, `config/parser.rs` and then `pub(super)` or completely private might also work? 

The current `config.rs` will move to `config/mod.rs` . This way later all the individual configs may be moved to their own modules. eg. `config/memory.rs` etc.

gabhijit (commented):
Mostly looks good. If you want to consider the visibility, otherwise this is quite good.

src/vmm/src/arg_parser.rs
FedericoPonzi: Good idea!

andreeaflorescu (approved):

lauralt (approved):

PR #57, opened 2021-01-13
update documentation

lauralt:


resources/make_test_resources.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
```
andreeaflorescu, line 1: Can we reuse this in the deps-pipeline so that we do not have duplicated commands?

docs/DESIGN.md
```diff
@@ -21,30 +21,28 @@ available in `rust-vmm`:
...
 
 Complex functionalities that require ongoing user interaction with the VMM are
-not supported in this version. Such functionalities include but are not
-limited to: device hotplug, VM pause/resume, snapshotting, migration. As the
-necessary building blocks emerge in `rust-vmm`, integration in the `vmm` crate
-will not pose a large effort, but the frontend / app will require extensive
-modifications.
+not currently supported. Such functionalities include but are not limited to:
+device hotplug, VM pause/resume, snapshotting, migration. We plan on adding the
+VM pause/resume support in the near future. As the necessary building blocks
```
andreeaflorescu, line 30: I've created an issue: https://github.com/rust-vmm/vmm-reference/issues/58. Can you mention the issue instead? Saying "near future" is a bit odd IMHO.

lauralt: Yup, done

resources/make_test_resources.sh
lauralt: Synced offline. The disk step requires `privileged: true`, so we can't reuse this script.

andreeaflorescu (dismissed):

README.md
```diff
@@ -95,6 +99,29 @@ vmm-reference                           \
...
+image format, i.e. ELF and bzImage) exercise the only public function of the
+`Vmm` object, `run()`.
+The Python integration tests make use of the VMM in varied configurations that
+aren’t overly complex and illustrate realistic use cases (e.g. one test runs a
+VMM with a virtio block device, one test will run a VMM with PCI, etc.).
+
+To be able to successfully run all the tests in this repo, there are a couple of
+resources that need to be found in `/tmp`. To generate the resources at the
+expected locations, you can run the following command:
+`sudo ./../resources/make_test_resources.sh`.
```
FedericoPonzi, line 118: You can also add that you need to `cd resources` first and then run this command from there. I would also like to see how to run the tests (via `cargo test`).

lauralt: Oops, I thought I'm in `docs`, fixed! Also added more details about how to run the tests, thanks for the feedback!

README.md
```diff
@@ -95,6 +99,64 @@ vmm-reference                           \
...
+```
+
+There is no single command yet for running all the Python integration tests in
+one shot. To run the tests from a single file, you can use:
+
+```bash
+pytest-3 <path_to_the_file>
+```
+For example:
+```bash
```
aghecenco, line 140: nit: surround code snippets with blank lines (here and below)

docs/DESIGN.md
```diff
@@ -21,30 +21,30 @@ available in `rust-vmm`:
...
 will not pose a large effort, but the frontend / app will require extensive
-modifications.
+modifications, see
+[tracking issue](https://github.com/rust-vmm/vmm-reference/issues/55).
 
 ## Overview
 
 The reference VMM is composed of 2 major crates packaged in a binary
-application: the `vmm` and a oneshot CLI that invokes it. The `vmm` crate
+application: the `vmm` and an oneshot CLI that invokes it. The `vmm` crate
```
aghecenco, line 40: nit: `a oneshot` is correct here, pronunciation-wise `oneshot` *sounds* as if it started with the consonant `w` (see [a more compelling explanation](https://www.grammar.com/a-vs-an-when-to-use))

resources/make_test_resources.sh
```diff
@@ -0,0 +1,23 @@
...
+# This script illustrates the build steps for all the kernel and disk images
+# used by tests in vmm-reference.
+
+set -e
+
+SOURCE=$(readlink -f "$0")
+TEST_RESOURCE_DIR="$(dirname "$SOURCE")"
+cd $TEST_RESOURCE_DIR
+
+./kernel/make_kernel_busybox_image.sh -f elf -k vmlinux-hello-busybox -w /tmp/vmlinux_busybox -j 2
```
aghecenco, line 15: nit: you could set variables with the various locations in `tmp` so that they're easier to modify; you could also check if the variable is already set and only modify it if it's unset (so that users who want different locations for the artifacts can do something like `KERN_DIR=/home/user/kernel ./make_test_resources.sh`)

aghecenco (dismissed):

docs/DESIGN.md
lauralt: :O Just followed the IDE suggestion :(. I'll revert, thanks!

andreeaflorescu (approved):

resources/make_test_resources.sh
lauralt: done

andreeaflorescu (approved):

README.md
aghecenco: Yep it's a not-exactly-official guideline from [here](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md031---fenced-code-blocks-should-be-surrounded-by-blank-lines), and also one of the things the [`mdl` markdown linter](https://github.com/markdownlint/markdownlint) complains about

aghecenco (approved):

PR #59, opened 2021-01-16
Implementation of Builder pattern for Config

gabhijit:
Implemented `Builder` for `VMMConfig`. (Note: This is an early
implementation, more like prototype. But almost all functionality is
present.)

Removed `derive(Default)` from `MemoryConfig`, `KernelConfig` and
`VcpuConfig` and added sensible defaults (memory = 256M, vcpu=1) (derive
default will trivially initialize this to 0 and hence we are required to
take a memory coniguration from CLI launcher).

Brought these changes to `api` (`CLI`) to demonstrate a use.

Removed `required` from 'memory' and 'vcpu' config.

Changed `TryFrom<String>` to `TryFrom<&str>`. Believe this is better
because it's quite possible to get `&str` from `String` than other way
round and in fact often we'll get `&str` from CLI parsing etc. But the
other way round is painful.

Right now passes all 'existing' tests (`cargo test --all`, except one
which likely fails because the right kernel image is not available to
test.)

TODO: Documentation with examples for using the `Builder` and Test cases
for the `Builder`.

### DIFF
--- a/src/api/src/lib.rs
+++ b/src/api/src/lib.rs
@@ -70,7 +70,7 @@ impl CLI {
             .memory_config(matches.value_of("memory"))
             .kernel_config(matches.value_of("kernel"))
             .vcpu_config(matches.value_of("vcpu"))
-            .network_config(matches.value_of("net"))
+            .net_config(matches.value_of("net"))
             .block_config(matches.value_of("block"))
             .build()
             .map_err(|e| format!("{:?}", e))?)
@@ -233,7 +233,7 @@ mod tests {
                 memory_config: MemoryConfig { size_mib: 128 },
                 vcpu_config: VcpuConfig { num: 1 },
                 block_config: None,
-                network_config: None,
+                net_config: None,
             }
         );
 
@@ -249,7 +249,7 @@ mod tests {
                 memory_config: MemoryConfig { size_mib: 256 },
                 vcpu_config: VcpuConfig { num: 1 },
                 block_config: None,
-                network_config: None,
+                net_config: None,
             }
         );
     }
--- a/src/api/src/lib.rs
+++ b/src/api/src/lib.rs
@@ -6,11 +6,10 @@
 //! CLI for the Reference VMM.
 
 #![deny(missing_docs)]
-use std::convert::TryFrom;
 use std::result;
 
 use clap::{App, Arg};
-use vmm::{BlockConfig, KernelConfig, MemoryConfig, NetConfig, VMMConfig, VcpuConfig};
+use vmm::VMMConfig;
 
 /// Command line parser.
 pub struct CLI;
@@ -26,17 +25,13 @@ impl CLI {
             .arg(
                 Arg::with_name("memory")
                     .long("memory")
-                    .required(true)
                     .takes_value(true)
-                    .default_value("size_mib=256")
                     .help("Guest memory configuration.\n\tFormat: \"size_mib=<u32>\""),
             )
             .arg(
                 Arg::with_name("vcpu")
                     .long("vcpu")
-                    .required(true)
                     .takes_value(true)
-                    .default_value("num=1")
                     .help("vCPU configuration.\n\tFormat: \"num=<u8>\""),
             )
             .arg(
@@ -71,52 +66,14 @@ impl CLI {
             format!("Invalid command line arguments: {}", e)
         })?;
 
-        Ok(VMMConfig {
-            memory_config: MemoryConfig::try_from(
-                matches
-                    .value_of("memory")
-                    .expect("Missing memory configuration")
-                    .to_string(),
-            )
-            .map_err(|e| {
-                eprintln!("{}", help_msg);
-                format!("{}", e)
-            })?,
-            kernel_config: KernelConfig::try_from(
-                matches
-                    .value_of("kernel")
-                    .expect("Missing kernel configuration")
-                    .to_string(),
-            )
-            .map_err(|e| {
-                eprintln!("{}", help_msg);
-                format!("{}", e)
-            })?,
-            vcpu_config: VcpuConfig::try_from(
-                matches
-                    .value_of("vcpu")
-                    .expect("Missing vCPU configuration")
-                    .to_string(),
-            )
-            .map_err(|e| {
-                eprintln!("{}", help_msg);
-                format!("{}", e)
-            })?,
-            block_config: match matches.value_of("block") {
-                Some(value) => Some(BlockConfig::try_from(value.to_string()).map_err(|e| {
-                    eprintln!("{}", help_msg);
-                    format!("{}", e)
-                })?),
-                None => None,
-            },
-            network_config: match matches.value_of("net") {
-                Some(value) => Some(NetConfig::try_from(value.to_string()).map_err(|e| {
-                    eprintln!("{}", help_msg);
-                    format!("{}", e)
-                })?),
-                None => None,
-            },
-        })
+        Ok(VMMConfig::builder()
+            .memory_config(matches.value_of("memory"))
+            .kernel_config(matches.value_of("kernel"))
+            .vcpu_config(matches.value_of("vcpu"))
+            .network_config(matches.value_of("net"))
+            .block_config(matches.value_of("block"))
+            .build()
+            .map_err(|e| format!("{:?}", e))?)
     }
 }
 
@@ -126,6 +83,7 @@ mod tests {
 
     use std::path::PathBuf;
     use vmm::DEFAULT_KERNEL_CMDLINE;
+    use vmm::{KernelConfig, MemoryConfig, VcpuConfig};
 
     #[test]
     fn test_launch() {
--- a/src/vmm/src/config/builder.rs
+++ b/src/vmm/src/config/builder.rs
@@ -141,14 +141,14 @@ impl Builder {
     /// You can see example of how to use this function in [`Example` section from
     /// `build`](#method.build)
     ///
-    pub fn network_config<T>(self, network: Option<T>) -> Self
+    pub fn net_config<T>(self, net: Option<T>) -> Self
     where
         NetConfig: TryFrom<T>,
         <NetConfig as TryFrom<T>>::Error: Into<ConversionError>,
     {
-        match network {
+        match net {
             Some(n) => self.and_then(|mut config| {
-                config.network_config = Some(TryFrom::try_from(n).map_err(Into::into)?);
+                config.net_config = Some(TryFrom::try_from(n).map_err(Into::into)?);
                 Ok(config)
             }),
             None => self,
@@ -274,22 +274,22 @@ mod tests {
     #[test]
     fn test_builder_net_config_none_default() {
         let vmm_config = Builder::default()
-            .network_config(None as Option<&str>)
+            .net_config(None as Option<&str>)
             .kernel_config(Some("path=bzImage"))
             .build();
         assert!(vmm_config.is_ok());
-        assert!(vmm_config.unwrap().network_config.is_none());
+        assert!(vmm_config.unwrap().net_config.is_none());
     }
 
     #[test]
     fn test_builder_net_config_success() {
         let vmm_config = Builder::default()
-            .network_config(Some("tap=tap0"))
+            .net_config(Some("tap=tap0"))
             .kernel_config(Some("path=bzImage"))
             .build();
         assert!(vmm_config.is_ok());
         assert_eq!(
-            vmm_config.unwrap().network_config,
+            vmm_config.unwrap().net_config,
             Some(NetConfig {
                 tap_name: "tap0".to_string()
             })
--- a/src/vmm/src/config/builder.rs
+++ b/src/vmm/src/config/builder.rs
@@ -1,4 +1,4 @@
-// Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
 //! Config builder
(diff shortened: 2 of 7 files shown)


gabhijit:
A prototype pull request for the Builder pattern for config request. This does not affect the `config`  refactoring of #55.

andreeaflorescu (commented):
I personally like the approach, I have a few comments on the code level, but I'll add them once other people chime in to see if we should move forward with this.

@lauralt @aghecenco @alexandruag what do you say?

The only thing that I find a bit out of hand is that in the builder we need to provide a default for `KernelConfig`. This could be avoided with some duplication, but we need to see if it is really important.

lauralt:
> I personally like the approach, I have a few comments on the code level, but I'll add them once other people chime in to see if we should move forward with this.
> 
> @lauralt @aghecenco @alexandruag what do you say?

I like the proposal too.
 
> The only thing that I find a bit out of hand is that in the builder we need to provide a default for `KernelConfig`. This could be avoided with some duplication, but we need to see if it is really important.

Would be nice if we can avoid the `Default` impl for `KernelConfig`, and use the `cmdline` and `himem_start` default values somewhere else, but if it becomes weird we can also open an issue for it and not block this PR with that (from what it looks like we are still required to pass the `path` arg and a value for it, am I right?).

src/vmm/src/config/builder.rs
```diff
@@ -0,0 +1,110 @@
...
+
+use super::{
+    BlockConfig, ConfigResult, ConversionError, KernelConfig, MemoryConfig, NetConfig, VMMConfig,
+    VcpuConfig,
+};
+
+/// Builder structure for VMMConfig
+#[derive(Debug)]
+pub struct Builder {
+    inner: ConfigResult<VMMConfig>,
```
andreeaflorescu, line 12: I find it a bit more readable without the `ConfigResult` type definition. When first reading the code, I thought that `ConfigResult` is a fancy type, and not just a wrapper over Result.

If you would like to keep the type, can you move it to this module instead?

src/vmm/src/config/builder.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+impl Builder {
+    pub fn new() -> Self {
+        Builder {
+            inner: Ok(VMMConfig::default()),
+        }
+    }
+
+    pub fn build(&self) -> ConfigResult<VMMConfig> {
```
andreeaflorescu, line 28: I've given it a bit more thought, and I don't think that there is really a nice way of disallowing defaults for `KernelConfig`. But, there is a hacky way :laughing: 

How about: we return an error here on `build`, when the path of the `KernelConfig` is empty? This would allow us to remove the `Default` implementation for `KernelConfig`, as well as returning an error in the builder as well when the required value is not set. Since this is not necessarily the nicest way of doing it, it would require documentation to explain why this hack is needed.

src/vmm/src/config/mod.rs
```diff
@@ -119,10 +144,20 @@ pub struct KernelConfig {
     pub himem_start: u64,
 }
 
-impl TryFrom<String> for KernelConfig {
+impl Default for KernelConfig {
```
andreeaflorescu, line 147: I would prefer to remove this so that people do not use by mistake. `KernelConfig` should not support a default.

src/vmm/src/config/mod.rs
```diff
@@ -60,16 +71,24 @@ impl fmt::Display for ConversionError {
...
 pub struct MemoryConfig {
     /// Guest memory size in MiB.
     pub size_mib: u32,
 }
 
-impl TryFrom<String> for MemoryConfig {
+impl Default for MemoryConfig {
+    fn default() -> Self {
+        MemoryConfig {
+            size_mib: 256 as u32,
```
andreeaflorescu, line 83: nit: you can use `256u32` instead so that you do not need the conversion.

src/vmm/src/config/mod.rs
```diff
@@ -85,16 +104,22 @@ impl TryFrom<String> for MemoryConfig {
...
+#[derive(Clone, Debug, PartialEq)]
 pub struct VcpuConfig {
     /// Number of vCPUs.
     pub num: u8,
 }
 
-impl TryFrom<String> for VcpuConfig {
+impl Default for VcpuConfig {
+    fn default() -> Self {
+        VcpuConfig { num: 1 as u8 }
```
andreeaflorescu, line 115: nit: you can use `1u8` instead.

src/vmm/src/config/mod.rs
```diff
@@ -153,16 +188,16 @@ impl TryFrom<String> for KernelConfig {
...
 /// Network device configuration.
-#[derive(Clone, Debug, Default, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct NetConfig {
     /// Name of tap device.
     pub tap_name: String,
 }
 
-impl TryFrom<String> for NetConfig {
+impl TryFrom<&str> for NetConfig {
```
andreeaflorescu, line 197: Using `&str` instead of `String` is a really nice change, good catch! It would be nice to have it as a separate commit since it is not directly related to the Builder. If this is too troublesome, there is no problem. We generally try to keep one change per commit because this makes it easier to do reverts when they're necessary. It is also easier to write the changelog because you do not have to look at each commit in depth to see if there are additional changes that the issue it resolves.

src/vmm/src/config/builder.rs
gabhijit: Ok Sure. I placed it inside `config/mod.rs`, because I thought it's generally going to be useful for all `config` related activities (like parser etc.), May be `ConfigResult` is a bad name :-)  For now will move to a simple `Result`.

src/vmm/src/config/builder.rs
gabhijit: When you say  remove the `Default` implementation, Do you also mean to say remove `[derive(Default)]` as well. That would make things complicated, because then we cannot use `derive(Default)` on the top level `VMMConfig`. 

I like the idea of throwing an `error` upon 'empty' kernel-config from the `build` function, but removing `Default` and/or `derive(Default)` would mean we cannot use `Default` for the entire `VMMConfig`, which would make everything mandatory (a bit of a UX annoyance) 

I believe this `build` should be a catchall place for all configuration errors by asserting values here, because this will indeed be the last in the chain (and a required one). Also, in the context of `clap` based CLI, this is not a huge problem, because missing kernel config will be reported as an error already, even before it comes here.

src/vmm/src/config/mod.rs
gabhijit: sure.

src/vmm/src/config/mod.rs
gabhijit: sure.

src/vmm/src/config/mod.rs
gabhijit: Yes, will do that - I was thinking of that already because regardless of whether `builder` is to be used, this is a good change. So may be I will first make that change and then look at other changes.

src/api/src/lib.rs
```diff
@@ -71,52 +66,14 @@ impl CLI {
...
-                    format!("{}", e)
-                })?),
-                None => None,
-            },
-        })
+        Ok(VMMConfig::builder()
+            .memory_config(matches.value_of("memory"))
+            .kernel_config(matches.value_of("kernel"))
+            .vcpu_config(matches.value_of("vcpu"))
+            .network_config(matches.value_of("network"))
```
lauralt, line 73: "network" should be "net". The sad thing is there was no test to catch this thing :( (sorry for that). Is it possible to add a test for this or include it in some other one (for example [this test](https://github.com/rust-vmm/vmm-reference/blob/master/src/api/src/lib.rs#L257))?

src/vmm/src/config/builder.rs
```diff
@@ -0,0 +1,122 @@
...
+        self.inner.clone()
+    }
+
+    pub fn memory_config<T>(self, memory: Option<T>) -> Self
+    where
+        MemoryConfig: TryFrom<T>,
+        <MemoryConfig as TryFrom<T>>::Error: Into<ConversionError>,
+    {
+        match memory {
+            Some(m) => self.and_then(move |mut config| {
```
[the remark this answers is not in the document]
gabhijit: I am not sure! :-) I tried it without `move` also, but the compiler didn't crib. TBH: I didn't understand, why. My earlier thinking was `move` here was necessary to consume the value, but compiler is just fine without `move`, so not sure. If you suggest `move` is not required. I'll remove it.

src/vmm/src/config/builder.rs
```diff
@@ -0,0 +1,122 @@
+//! Config builder
```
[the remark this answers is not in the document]
gabhijit: Sure

src/api/src/lib.rs
gabhijit: Oh. Good catch! Fixed.

src/vmm/src/config/builder.rs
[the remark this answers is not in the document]
gabhijit: Well - I get it now, since we are not using the closure beyond that function `move` is redundant.

src/api/src/lib.rs
gabhijit: I have added a separate commit that consistently uses `net_config`. Hope this is fine, if this is not required, We can simply revert that commit.

lauralt: Looks good to me; the commit message is a bit misleading, I think it should've been: We were using the network config param as **`network_config`**...

Regarding the other changes, I'll have a look on Monday. Thanks!

src/vmm/src/config/builder.rs
```diff
@@ -0,0 +1,375 @@
...
+            }
+            Err(_) => {}
+        }
+
+        self.inner.clone()
+    }
+
+    /// Configure Builder with Memory Configuration for the VMM
+    ///
+    /// # Example
```
lauralt, line 71: Just a quick comment, if the examples turn out to be very similar (as it happens in this module), I would suggest to have a single one at the module level that uses all the public functions, and in the `Example` section of each function to point to that one.

You can find a hopefully decent example with this approach in [this module](https://github.com/rust-vmm/vm-superio/blob/master/src/i8042.rs).

src/vmm/src/config/builder.rs
```diff
@@ -123,3 +237,139 @@ impl Builder {
...
+        let vmm_config = Builder::default()
+            .block_config(None as Option<&str>)
+            .kernel_config(Some("path=bzImage"))
+            .build();
+        assert!(vmm_config.is_ok());
+        assert!(vmm_config.unwrap().block_config.is_none());
+    }
+
+    #[test]
+    fn test_builder_block_config_success() {
```
lauralt, line 362: I would also add a test with all the configurations done, and check that `vmm_config` is the expected one, e.g.

```rust
fn test_builder_vmm_config_success() {
        let vmm_config = Builder::default()
            .memory_config(Some("size_mib=1024"))
            .vcpu_config(Some("num=2"))
            .net_config(Some("tap=tap0"))
            .kernel_config(Some("path=bzImage"))
            .block_config(Some("path=/dev/loop0"))
            .build();
        assert!(vmm_config.is_ok());
        assert_eq!(
            vmm_config.unwrap(),
            VMMConfig {
                memory_config: MemoryConfig { size_mib: 1024 },
                vcpu_config: VcpuConfig { num: 2 },
                kernel_config: KernelConfig {
                    cmdline: DEFAULT_KERNEL_CMDLINE.to_string(),
                    himem_start: DEFAULT_HIGH_RAM_START,
                    path: PathBuf::from("bzImage")
                },
                net_config: Some(NetConfig {
                    tap_name: "tap0".to_string()
                }),
                block_config: Some(BlockConfig {
                    path: PathBuf::from("/dev/loop0")
                })
            }
        );
}
```

src/vmm/src/config/mod.rs
```diff
@@ -66,10 +66,18 @@ pub struct MemoryConfig {
     pub size_mib: u32,
 }
 
-impl TryFrom<String> for MemoryConfig {
+impl Default for MemoryConfig {
```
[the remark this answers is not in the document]
gabhijit: This is a bit difficult.. What I can do is. Delete this `impl Default` create a new commit and then squash into next one. The normal `git diff --patch` won't  let me split the patch! Even I didn't like it so. But I was like - fine. But let me see if it's not overly annoying to do.

src/vmm/src/config/builder.rs
```diff
@@ -123,3 +237,139 @@ impl Builder {
...
+        assert!(vmm_config.unwrap().network_config.is_none());
+    }
+
+    #[test]
+    fn test_builder_net_config_success() {
+        let vmm_config = Builder::default()
+            .network_config(Some("tap=tap0"))
+            .kernel_config(Some("path=bzImage"))
+            .build();
+        assert!(vmm_config.is_ok(), "{:?}", vmm_config.err());
```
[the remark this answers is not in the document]
gabhijit: Leftovers from the debugging! :-) Good Catch! For now I will not print it, but in general I think it's not a bad idea to have these prints, but there are many places where they should be added (and consistently!)

src/vmm/src/config/builder.rs
gabhijit: Oh!! Thanks! Totally missed this.

src/vmm/src/config/builder.rs
```diff
@@ -16,17 +16,43 @@ pub struct Builder {
...
-        }
+        Builder::default()
     }
 
+    /// Builds VMMConfig
+    ///
+    /// This function should be last of the chain to build VMMConfiguration. Since, it is not known
+    /// which of the `config` functions user has called, missing configuration errors are handled
+    /// here -
+    ///  * Empty Kernel Configuration ('path' missing.)
```
andreeaflorescu, line 37: This comment reveals implementation details. How about we just keep the format typically used by Rust crates and have a short description, then an error section?

For example:
```rust
/// Builds a `VMMConfig`. This function needs to be called after all needed configurations are set.
///
/// # Errors
///
/// Returns an error when the required fields were not set using the appropriate `*_config` functions.
```

This way we do not explicitly mention the required fields, which means that if we add other in the future, this comment will not be outdated.

src/vmm/src/config/builder.rs
```diff
@@ -0,0 +1,354 @@
...
+}
+
+impl Builder {
+    /// Creates a Builder Object
+    ///
+    pub fn new() -> Self {
+        Builder::default()
+    }
+
+    /// Builds VMMConfig
```
andreeaflorescu, line 32: For consistency reasons, can you please end all sentences in documentation with `.`?

andreeaflorescu (commented):
Just a couple of nits. The code looks good to me :+1:

src/vmm/src/config/builder.rs
gabhijit: I believe it's not possible to get rid of `Default` implementation for `KernelConfig` and hence at-least for now we are going ahead with catching errors in the `build` function as implemented. Hence marking this conversation as resolved.

src/vmm/src/config/mod.rs
```diff
@@ -11,10 +11,14 @@ use arg_parser::CfgArgParser;
 
 use super::{DEFAULT_HIGH_RAM_START, DEFAULT_KERNEL_CMDLINE};
 
+use builder::Builder;
+
 mod arg_parser;
 
+mod builder;
```
lauralt, line 18: Hmm, unless I'm missing something here, to be able to see the `builder` in the generated documentation, I think this should be `pub mod` (or `pub use builder::Builder` above for just the `Builder` functionality).

lauralt (approved):

andreeaflorescu (approved):

Issue #61, opened 2021-02-02
Issue with building test images on Ubuntu 20.04 


gabhijit:
I am trying to build test images on Ubuntu 20.04, by running the following commands - (taken from `README.md`) 

```
sudo ./resources/make_test_resources.sh
```
I am getting following error. 

```shell
/usr/bin/ld: util-linux/lib.a(rdate.o): in function `rdate_main':
rdate.c:(.text.rdate_main+0x11e): undefined reference to `stime'
/usr/bin/ld: coreutils/lib.a(date.o): in function `date_main':
date.c:(.text.date_main+0x27e): undefined reference to `stime'
collect2: error: ld returned 1 exit status
Note: if build needs additional libraries, put them in CONFIG_EXTRA_LDLIBS.
Example: CONFIG_EXTRA_LDLIBS="pthread dl tirpc audit pam"
make: *** [Makefile:717: busybox_unstripped] Error 1

```
There are a bunch of warnings too (not captured above). The error primarily comes because the `stime`appears to be deprecated in  `libc` version is 2.31 (default on Ubuntu 20.04). 

This can likely be fixed by using a different version of busybox (Busybox 1.32). 

Or is there another alternative?

andreeaflorescu:
> I am trying to build test images on Ubuntu 20.04, by running the following commands - (taken from `README.md`)
> 
> ```
> sudo ./resources/make_test_resources.sh
> ```
> 
> I am getting following error.
> 
> ```shell
> /usr/bin/ld: util-linux/lib.a(rdate.o): in function `rdate_main':
> rdate.c:(.text.rdate_main+0x11e): undefined reference to `stime'
> /usr/bin/ld: coreutils/lib.a(date.o): in function `date_main':
> date.c:(.text.date_main+0x27e): undefined reference to `stime'
> collect2: error: ld returned 1 exit status
> Note: if build needs additional libraries, put them in CONFIG_EXTRA_LDLIBS.
> Example: CONFIG_EXTRA_LDLIBS="pthread dl tirpc audit pam"
> make: *** [Makefile:717: busybox_unstripped] Error 1
> ```
> 

@gabhijit thanks for reporting this issue.

> There are a bunch of warnings too (not captured above). The error primarily comes because the `stime`appears to be deprecated in `libc` version is 2.31 (default on Ubuntu 20.04).
> 

[...]

PR #62, opened 2021-02-04
resources: build busybox on glibc 2.31+ (#61)

gabhijit:
busybox 1.26 does not build if the host glibc version is 2.31 this is
due to `stime` call now not supported. The corresponding fix for the
busybox is available in busybox 1.32. We are using the current stable
version for 1.32 which is 1.32.1. The config is abstracted out into a
separate file 'configs.sh' which is sourced by the respective
`make_*.sh` file. Current `glibc` version is determined using a
mechanism similar to `configure` scripts. Small readability changes
inside `make_busybox.sh`(added empty lines to keep actions separate).
Trailing white spaces removed in make_kernel.sh.


resources/kernel/configs.sh
```diff
@@ -0,0 +1,41 @@
...
+GLIBC_VER=`gcc -o glibc_ver glibc_ver.c 2>/dev/null && ./glibc_ver 2>/dev/null`
+
+#Delete the files generated
+rm -f glibc_ver* 2> /dev/null
+
+if [[ -z $GLIBC_VER ]]; then
+	echo "Unable to determine glibc version, using default Busybox: 1.26."
+	BUSYBOX_CFG="busybox_1_26_static_config"
+	BUSYBOX_VERSION="1.26.0"
+else
```
aghecenco, line 30: This part can be omitted for brevity; if `$GLIBC_VER` is not set, `[[ $GLIBC_VER < 2.31 ]]` evaluates to `true`

resources/kernel/configs.sh
```diff
@@ -0,0 +1,41 @@
...
+		BUSYBOX_CFG="busybox_1_26_static_config"
+		BUSYBOX_VERSION="1.26.0"
+	else
+		BUSYBOX_CFG="busybox_1_32_1_static_config"
+		BUSYBOX_VERSION="1.32.1"
+	fi
+
+fi
+
+echo "Busybox Version: $BUSYBOX_VERSION Config: $BUSYBOX_CFG"
```
aghecenco, line 41: nit: I'd split this into 2 separate `echo`s so that `Config: ...` is on a separate line

resources/kernel/configs.sh
gabhijit: This part I kept separate because, if this gets logged through redirect (`> 1 2>&1 `), it will help in debugging why a particular version was selected, so this becomes explicit, but without that it's implicit.

resources/kernel/configs.sh
gabhijit: Sure, will fix it. In fact I believe only `config` is sufficient. I added the version `echo` just to make sure somewhere it's not accidentally forgotten (did happen once while testing it! :-) )

resources/kernel/configs.sh
gabhijit: Also, if in future one has to make `1.32` as a default (likely as more and more distros move to `glibc 2.31+`), relying on the above evaluation to `true` won't help. So I am not entirely sure it's a good idea to omit this part.

andreeaflorescu (approved):

aghecenco (approved):

Issue #63, opened 2021-02-05
Small Improvements to integration tests


gabhijit:
Suggesting a couple of improvements to integration tests inside `tests/test_run_reference_vmm.py` 

1. This one is more like a 'nit'  

```python 
def start_vmm_process(kernel_path, disk_path, num_vcpus = 1, mem_size_mib = 1024):
   ...
```
In the function above, `disk_path` parameter should be optional, then it will also be consistent with other semantics where only kernel path is a required parameter while configuring the VM. 

2. This one is more like a bug - 

```python 

def expect_string(vmm_process, expected_string):
    while True:
        nextline = vmm_process.stdout.readline()
        if expected_string in nextline.decode():
 ...

```

The `while True` loop above does not break and  if the case is failing, it just remains stuck (Observed it a couple of times while running tests that failed). The only option is to `kill` the Python process. This needs to be handled with `subprocess.wait` - Which throws `TimeoutExpired` and catch it around some number of retries and fail the test. It is safe to catch the exception and retry [as discussed here](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait)

andreeaflorescu:
> 1. This one is more like a bug -
> 
> ```python
> def expect_string(vmm_process, expected_string):
>     while True:
>         nextline = vmm_process.stdout.readline()
>         if expected_string in nextline.decode():
>  ...
> ```
> 
> The `while True` loop above does not break and if the case is failing, it just remains stuck (Observed it a couple of times while running tests that failed). The only option is to `kill` the Python process. This needs to be handled with `subprocess.wait` - Which throws `TimeoutExpired` and catch it around some number of retries and fail the test. It is safe to catch the exception and retry [as discussed here](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait)

This one is already reported: https://github.com/rust-vmm/vmm-reference/issues/18
We were thinking of fixing it by using pytest-timeout, but we did not install it in the container yet. I'll publish a new container version, and then install `pytest-timeout` as well, but we can also fix it as you suggested. Are you interested in submitting a PR with your suggestion?

lauralt:
@gabhijit I updated the issue description with the explicit code, linking a file line was not a good idea.

gabhijit:
> @gabhijit I updated the issue description with the explicit code, linking a file line was not a good idea.

[...]

Issue #46, opened 2020-12-14
Investigate long boot wait time in block device test


alexandruag:
Figure out why the `(bzimage-focal, rootfs.ext4)` combination takes too long to boot for `test_reference_vmm_with_disk` in `tests/test_run_reference_vmm.py` (see comment above `KERNELS_DISK` global).

gabhijit:
I have been able to look at this (or actually ended up looking at this in the context of something else!). Here is what happens, 

1.  When we use `bzimage-focal` image, the root filesystem is mounted as **read-only**. See below from the output of `dmesg` - 
```
....
[    0.622651] Loaded X.509 cert 'Build time autogenerated kernel key: e885ede42
fd054dd93a39894f24e3d3fcc02822c'
[    0.623279] zswap: default zpool zbud not available
[    0.623617] zswap: pool creation failed
[    0.623928] Key type ._fscrypt registered
[    0.624196] Key type .fscrypt registered
[    0.624485] Key type encrypted registered
[    0.625917] EXT4-fs (vda): mounted filesystem with ordered data mode. Opts: (null)
[    0.626366] VFS: Mounted root (ext4 filesystem) readonly on device 254:0.
....
```

Consequently a number of processes that rely on being able to write to FS fail. `systemd-logind`, `systemd-timesynd`, `systemd-resolved` to name a few.  Oddly enough the login shell is provided. Don't know why? 

2. When we use `vmlinux-focal` image, root FS is mounted **read-write** 

```
 ...
[    0.647635] Loading compiled-in X.509 certificates
[    0.648463] Loaded X.509 cert 'Build time autogenerated kernel key: e885ede42
fd054dd93a39894f24e3d3fcc02822c'
[    0.649155] zswap: default zpool zbud not available
[    0.649485] zswap: pool creation failed
[    0.649797] Key type ._fscrypt registered
[    0.650084] Key type .fscrypt registered
[    0.650414] Key type encrypted registered
[    0.788979] EXT4-fs (vda): mounted filesystem with ordered data mode. Opts: (null)
[    0.792337] VFS: Mounted root (ext4 filesystem) on device 254:0.
....

```

All the services are started properly and hence we quickly get a login shell. 

The root cause appears to be the read-only mount of the root filesystem. Why that happens with `bzimage`, I don't know! :-(

gabhijit:
[...]

Issue #18, opened 2020-10-30
add pytest timeout to test_run_reference_vmm.py


lauralt:
If the boot process in test_run_reference_vmm.py hangs here:
```rust
while True:
    nextline = vmm_process.stdout.readline()
    if expected_string in nextline.decode():
 ...
```
 the test fails in ci because of a [buildkite pipeline timeout](https://github.com/rust-vmm/vmm-reference/blob/master/.buildkite/custom-pipeline.yml#L15), but if we run the test locally, the test hangs too.
We can add a pytest timeout to the test when [this](https://github.com/rust-vmm/rust-vmm-container/issues/33) is unblocked or maybe find some other mechanism to fail the test after x time (is there a nice way to add a timeout to that `while`?).

PR #64, opened 2021-02-08
Integration tests improvements

gabhijit:
Contains fixes for the following -

#63 - improvements as discussed
#18 - Covered by improvements in #63
#46 - Long boot time for bzimage-focal

### DIFF
--- a/tests/test_run_reference_vmm.py
+++ b/tests/test_run_reference_vmm.py
@@ -13,13 +13,10 @@ KERNELS_INITRAMFS = [
     "/tmp/vmlinux_busybox/linux-4.14.176/bzimage-hello-busybox"
 ]
 
-"""
-Temporarily removed the ("bzimage-focal", "rootfs.ext4") pair from the
-list below, because the init startup sequence in the guest takes too
-long until getting to the cmdline prompt for some reason.
-"""
 KERNELS_DISK = [
     ("/tmp/ubuntu-focal/linux-5.4.81/vmlinux-focal",
+     "/tmp/ubuntu-focal-disk/rootfs.ext4"),
+    ("/tmp/ubuntu-focal/linux-5.4.81/bzimage-focal",
      "/tmp/ubuntu-focal-disk/rootfs.ext4"),
 ]
 
@@ -45,7 +42,7 @@ and run locally.
 
 def start_vmm_process(kernel_path, disk_path=None, num_vcpus=1, mem_size_mib=1024):
     # Kernel config
-    cmdline = "console=ttyS0 i8042.nokbd reboot=t panic=1 pci=off"
+    cmdline = "console=ttyS0 i8042.nokbd reboot=t panic=1 pci=off rw"
 
     himem_start = 1048576
 
@@ -55,7 +52,7 @@ def start_vmm_process(kernel_path, disk_path=None, num_vcpus=1, mem_size_mib=102
     vmm_cmd = [
         "target/release/vmm-reference",
         "--memory", "size_mib={}".format(mem_size_mib),
-        "--kernel", "cmdline=\"{}\",path={},himem_start={}".format(
+        "--kernel", "cmdline={},path={},himem_start={}".format(
             cmdline, kernel_path, himem_start
         ),
         "--vcpu", "num={}".format(num_vcpus)
@@ -135,7 +132,6 @@ def expect_string(vmm_process, expected_string, timeout=TEST_TIMEOUT):
             time.sleep(1)
             now = time.time()
             if now - then > giveup_after:
-            if now - then > giveup_after :
                 raise TimeoutError(
                         "Timed out {} waiting for {}".format(now - then, expected_string))
         except Exception as _:
--- a/tests/test_run_reference_vmm.py
+++ b/tests/test_run_reference_vmm.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 """Run the reference VMM and shut it down through a command on the serial."""
 
-import os, signal, subprocess, time
+import os, signal, subprocess, time, fcntl
 import pytest
 from subprocess import PIPE, STDOUT
 import tempfile
@@ -23,6 +23,8 @@ KERNELS_DISK = [
      "/tmp/ubuntu-focal-disk/rootfs.ext4"),
 ]
 
+# No. of seconds after which to give up for the test
+TEST_TIMEOUT = 30
 
 def process_exists(pid):
     try:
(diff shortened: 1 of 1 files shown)


tests/test_run_reference_vmm.py
```diff
@@ -93,33 +95,59 @@ def shutdown(vmm_process):
...
+            x = os.read(vmm_process.stdout.fileno(), 4096)
+            all_data += x
+            for line in all_data.split(b'\r\n'):
+                if expected_string in line.decode():
+                    return line.decode()
+                    found = True
+            # Whatever remains is collected in `all_data`.
+            all_data = line
+        except Exception as e:
+            time.sleep(1)
```
andreeaflorescu, line 129: Why do we sleep here?

tests/test_run_reference_vmm.py
```diff
@@ -93,33 +95,59 @@ def shutdown(vmm_process):
...
+
+    # No. of seconds after which we'll give up
+    giveup_after = timeout
+    then = time.time()
+    found = False
+    # This is required because the pattern we are expecting might get split across two reads
+    all_data = bytes()
+    while not found:
+        try:
+            x = os.read(vmm_process.stdout.fileno(), 4096)
```
andreeaflorescu, line 120: Is it possible to still use `readline`? It could make the code more readable than manually splitting by `\r\n`.

tests/test_run_reference_vmm.py
```diff
@@ -93,33 +95,59 @@ def shutdown(vmm_process):
     # TimeoutExpired exception and fail the test.
     vmm_process.wait(timeout=3)
 
+def setup_stdout_nonblocking(vmm_process):
```
andreeaflorescu, line 95: nit: can you add 2 newlines between all functions? That's the formatting that  PEP 8 expects. We do not yet have an integration test with that though :(

tests/test_run_reference_vmm.py
```diff
@@ -93,33 +95,59 @@ def shutdown(vmm_process):
...
+
+    ## We'll need to do non-blocking I/O with the underlying sub-process since
```
andreeaflorescu, line 97: nit: remove the extra #

tests/test_run_reference_vmm.py
```diff
@@ -93,33 +95,59 @@ def shutdown(vmm_process):
...
+        try:
+            x = os.read(vmm_process.stdout.fileno(), 4096)
+            all_data += x
+            for line in all_data.split(b'\r\n'):
+                if expected_string in line.decode():
+                    return line.decode()
+                    found = True
+            # Whatever remains is collected in `all_data`.
+            all_data = line
+        except Exception as e:
```
andreeaflorescu, line 128: What exception are we expecting here? Can we make it more explicit than `Exception`?

tests/test_run_reference_vmm.py
```diff
@@ -93,33 +95,59 @@ def shutdown(vmm_process):
...
+            time.sleep(1)
+            now = time.time()
+            if now - then > giveup_after :
+                raise TimeoutError("Timed out {} waiting for {}".format(now - then, expected_string))
```
andreeaflorescu, line 132: nit: this line is too long. PEP 8 expects lines to be maximum 100 chars long.

tests/test_run_reference_vmm.py
```diff
@@ -157,7 +199,7 @@ def test_reference_vmm(kernel):
     # If we do not find the expected message, this loop will not break and the
     # test will fail when the timeout expires.
     expected_string = "Hello, world, from the rust-vmm reference VMM!"
-    expect_string(vmm_process, expected_string)
+    _ = expect_string(vmm_process, expected_string)
```
andreeaflorescu, line 199: nit: Can you delete the `_ =`? Python seems to ignore the result by default so it's not really needed. Same for lines: 171 and 185.

andreeaflorescu (commented):
LGTM. I just have a few nits & questions.

tests/test_run_reference_vmm.py
gabhijit: The exception is [BlockingIOError](https://docs.python.org/3/library/exceptions.html#BlockingIOError). This is like `EWOULDBLOCK` so retrying immediately would likely throw the exception again, hence the idea is to sleep for a second and try. I could have done this with `select` and timeout as well.

tests/test_run_reference_vmm.py
gabhijit: Unfortunately no. I tried to use `stdout.readline` but it does not throw the `BlockingIOError` exception as above and simply blocks when it can't find a new-line. 

I agree this whole part is somewhat difficult to understand (if not very ugly), but unfortunately, I cannot think of a cleaner way of doing this. This happens because `Popen.communicate` closes the `stdin` that we want to use later, so effectively it means we have to do all the communication with the process by hand and hence all the trickery is required!

tests/test_run_reference_vmm.py
gabhijit: Okay sure.

tests/test_run_reference_vmm.py
gabhijit: Thanks, I added the explicit `BlockingIOError` exception and then removed it for something, may be leftover of debugging.

Will add two separate, we `retry` only on `BlockingIOError` and give up on any other `Exception`. That seems better.

tests/test_run_reference_vmm.py
gabhijit: Actually, in the past `pylint` has complained a lot when I didn't use output of a function that returned one, I am sure `pylint` or any of it's brethren would complain if we ignore.

tests/test_run_reference_vmm.py
andreeaflorescu: We can leave it as is, no problem.

tests/test_run_reference_vmm.py
andreeaflorescu: Ah, ok. I did not know that. We can then leave it as is, thanks!

gabhijit: Well, I tried using `pylint` to check if it indeed complains, it didn't, so I removed as you suggested.

tests/test_run_reference_vmm.py
```diff
@@ -93,33 +95,59 @@ def shutdown(vmm_process):
...
+            for line in all_data.split(b'\r\n'):
+                if expected_string in line.decode():
+                    return line.decode()
+                    found = True
+            # Whatever remains is collected in `all_data`.
+            all_data = line
+        except Exception as e:
+            time.sleep(1)
+            now = time.time()
+            if now - then > giveup_after :
```
andreeaflorescu, line 134: nit: extra space after `giveup_after`.

andreeaflorescu (approved):

tests/test_run_reference_vmm.py
```diff
@@ -2,27 +2,26 @@
...
-"""
 KERNELS_DISK = [
     ("/tmp/ubuntu-focal/linux-5.4.81/vmlinux-focal",
      "/tmp/ubuntu-focal-disk/rootfs.ext4"),
+    ("/tmp/ubuntu-focal/linux-5.4.81/bzimage-focal",
+     "/tmp/ubuntu-focal-disk/rootfs.ext4"),
 ]
 
+# No. of seconds after which to give up for the test
+TEST_TIMEOUT = 60
```
lauralt, line 24: Isn't this timeout a bit too much?

lauralt (approved):

tests/test_run_reference_vmm.py
gabhijit: Should I make it 30 seconds,? I will have to rebase the commits once, so I can make it 30 seconds, if that's fine.

andreeaflorescu (approved):

lauralt (dismissed):

tests/test_run_reference_vmm.py
```diff
@@ -94,32 +93,66 @@ def shutdown(vmm_process):
...
+                    found = True
+                    return line.decode()
+            # Whatever remains is collected in `all_data`.
+            all_data = line
+        except BlockingIOError as _:
+            # Raised on `EWOULDBLOCK`, so it's better to wait for sometime before retrying
+            time.sleep(1)
+            now = time.time()
+            if now - then > giveup_after:
+            if now - then > giveup_after :
```
lauralt, line 135: this should be removed

lauralt (approved):

tests/test_run_reference_vmm.py
gabhijit: yeah.. sorry!!! some goof up during rebasing! :-( Yikes!

andreeaflorescu (approved):

Issue #47, opened 2020-12-14
Improve block device CI test


alexandruag:
Run something like `lsblk --json` and check output data such as device name and read-only status. Original proposal in [this](https://github.com/rust-vmm/vmm-reference/pull/41/commits/ac4a9dd45a7b8ce1d299634db3bb96d2a249d483#r540970103) comment.

gabhijit:
I can implement this if no one is taking this. 

I am wondering if it's a good idea to implement by hand a functionality equivalent to `Popen.communicate` while trying to implement fix for this. 

Right now with `expect_string` we can execute only one command (by sending a `bytestring` to the `stdin` like we are doing in `expect_vcpu`), but for executing more than one commands current `expect_string` alone won't work. So something like 

```
def run_cmd_in_vm(cmd_input, timeout):
    # returns the output of running the command or raise Exception.
``` 

Which can allow to run any number of commands once the VM is booted. 

@andreeaflorescu @lauralt what are your thoughts?

lauralt:
Yup, having some sort of abstraction for running a command was something that we were thinking about too. We would've also liked to have something that abstracts away the mechanism by which we send the command to the vm (e.g. stdin, ssh once we merge #49 ). Andreea had more ideas around this.

gabhijit:
I think with `ssh` (once the net functionality is in place), it's relatively easier, one can simply use something like [`paramiko`](https://www.paramiko.org/) (assuming an `sshd` runs inside the VM). 

I will send a PR that implements a basic abstraction for running command while and a fix for this issue.

andreeaflorescu:
We were considering using `pexect` for checking the output of commands. It would be nice to have an abstraction over the channel used for sending the commands because we would like to be able to test the vmm-reference with and without network which means that we cannot always count on the ssh capability to be available.

I am still thinking about the best abstraction, but I was looking for something as follows (we need better names):
-  define a `Channel` trait that has an `execute` command
- Implement the `Channel` trait for `SerialConsole` and for `Ssh`
- the communication `Channel` can be passed to the tests, which then use the `execute` function to check the expected behavior
- in the implementation of the SSH class we can use paramiko

With this approach we can run the same tests with the serial console, or with SSH which is useful for testing multiple flavors of the vmm.

gabhijit:
> 
> With this approach we can run the same tests with the serial console, or with SSH which is useful for testing multiple flavors of the vmm.

[...]

PR #66, opened 2021-02-16
tests: Added assertions for root  device  (#47)

gabhijit:
Added Assertions for root device 'name' and 'readonly' status. To be
able to do this, we need a mechanism to 'run' commands inside the VM.
Implemented a function `run_cmd_inside_vm` that runs the command and
returns the output of the command.

Signed-off-by: Abhijit Gadgil <gabhijit@iitbombay.org>


gabhijit:
This contains an implementation for `run_cmd_inside_vm`. I didn't use `pexpect` as it would have been a big change right from `start_vmm_process`. It's better to do that in a separate issue.

tests/test_run_reference_vmm.py
```diff
@@ -138,6 +139,43 @@ def expect_string(vmm_process, expected_string, timeout=TEST_TIMEOUT):
...
+    giveup_after = timeout
+    while True:
+        try:
+            data = os.read(vmm_process.stdout.fileno(), 4096)
+            output_lines = data.split(b'\r\n')
+            last = output_lines[-1].strip()
+            if prompt in last:
+                # FIXME: WE get the prompt twice in the output at the end,
+                # So removing it. No idea why twice?
+                # First one is 'cmd'
```
andreeaflorescu, line 166: You can get it once because you write to stdin, and once because it is captured by the serial console I think.

andreeaflorescu (approved):

tests/test_run_reference_vmm.py
```diff
@@ -214,12 +253,21 @@ def test_reference_vmm_with_disk(kernel, disk):
 
     vmm_process, tmp_disk_path = start_vmm_process(kernel, disk_path=disk)
 
-    expect_string(vmm_process, "Ubuntu 20.04 LTS ubuntu-rust-vmm ttyS0")
+    prompt = 'root@ubuntu-rust-vmm:~#'
+    expect_string(vmm_process, prompt)
+
+    cmd = 'lsblk --json'
+    output = run_cmd_inside_vm(cmd.encode(), vmm_process, prompt.encode(), timeout=5)
```
andreeaflorescu, line 260: Waiting for 5 seconds seems like too much. The output of `lsblk` should come instantly.

tests/test_run_reference_vmm.py
```diff
@@ -138,6 +139,43 @@ def expect_string(vmm_process, expected_string, timeout=TEST_TIMEOUT):
             raise
 
 
+def run_cmd_inside_vm(cmd, vmm_process, prompt, timeout=5):
```
andreeaflorescu, line 142: With adding this function we now use multiple ways of sending commands to the vmm. Is it possible to unify them? For example, by using this utility also in the tests that check the vcpu & memory?

tests/test_run_reference_vmm.py
gabhijit: Okay. I should change this to may be 1 sec. So for whatever reasons if `lsblk --json` does not return in 1 seconds that can be treated as failure right? I will also have to modify the `time.sleep` below to use something like `0.2` seconds, so that  we don't  get false positive exceptions.

tests/test_run_reference_vmm.py
gabhijit: Well - I am not sure why it is obtained twice at the end :-( But this too will go away with `pexpect`.

tests/test_run_reference_vmm.py
gabhijit: Actually, when we use `pexpect` a lot of this code is just going to go away (or will change substantially), so it's not a very good investment at the moment to clean that up.

I looked at the `pexpect` code which does something very similar to the above 'hand cooked' non-blocking (in a better and proven way!), so likely that's the path to be taken at some point of time soon enough.

andreeaflorescu: Hmmm, in this case I wonder if we need to go through this intermediate stage in which we use `run_cmd_inside_vm`. It can be confusing when reading the tests to see that some use one way of executing commands, and other use a different way.

gabhijit: The function would still be called `run_cmd_inside_vm` - except internally it will use `pexpect` instead of the `while True` loop. The function will still have (mostly) identical signature. But to be able to do that first we need to change the `create_vmm_process`, which I think is a big change and quite unrelated to this specific issue! (block device test).

Yes, initially my thoughts were same, these two ways are different and confusing. May be a good idea is to add a `FIXME` to those functions now and leave them as it is right now and fix them along with the implementation of `pexpect`. 

I will  try to work on a  PR using `pexpect` sometime next week.

tests/test_run_reference_vmm.py
```diff
@@ -138,6 +139,43 @@ def expect_string(vmm_process, expected_string, timeout=TEST_TIMEOUT):
...
+       Note: `cmd` and `prompt` should be a `bytes` object and not `str`.
+    """
+    cmd = cmd.strip() + b'\r\n'
+
+    vmm_process.stdin.write(cmd)
+    vmm_process.stdin.flush()
+
+    then = time.time()
+    giveup_after = timeout
+    while True:
```
aghecenco, line 158: I think this `while` loop can conceptually be replaced by overriding the timeout on each test that calls this function. Something like:

```python
@pytest.mark.timeout(10)
def test_reference_vmm_timeout():
  """ Times out in 10 seconds. """
  ...
  output = run_cmd_inside_vm(cmd.encode(), vmm_process, prompt.encode())
```

It'd be moving the timeout from per-command to per-test, but overall it's less verbose. WDYT?

gabhijit: I am not sure whether I am getting what you are trying to say. From what I understand you suggest, rather than having a `while True....`  we can just time out the test if whatever we expect to happen does not happen. 

It's a bit difficult to  achieve, because once we set the reading to non-blocking, we've to read in some kind of loop to read what we want and we cannot 'not' set it non-blocking (for the `expect_string`) part. 

The problem with this implementation is - there is a repetition of what we are doing with `epxect_string` and `run_cmd_inside_vm`. Ideally both this should be merged into a function that gives an API that says - read everything up to timeout or read up to expected string.  That's almost like an API that `pexpect` provides - more or less. 

In general, it's a good idea to start using the `pexpect` APIs but that change is substantial (discussed in other comments.) 

Having a per test timeout is a good idea - but I believe that is provided by `pytest-timeout` which is another package (and not available with `pytest` itself.) that is not there available with the CI docker Image IIRC. 

With `pexpect` (available in CI docker image) we get nicer APIs  (running commands) and get everything that `pytest-timeout` would achieve in this particular case. 

So this `while True` kludge will go away after `pexpect`.

aghecenco: Agreed - I tried a quick and dirty local refactor to see what it would look like with `pexpect` and it cleans up nicely. We can track adding `pytest-timeout` separately, it's not related to this improvement. Thanks for the explanation!

aghecenco (approved):

PR #67, opened 2021-02-18
Test Improvements

andreeaflorescu:
The main purpose of this PR is to provide the test resources from S3, instead of needing to build them (either locally or as part of the CI).

To provide these resources, we created a public S3 bucket that contains versioned resources, a manifest file that describes the resources, and the scripts committed to this repository under `resources`.

This PR is an RFC because it is missing documentation.


tests/tools/s3_download.py
```diff
@@ -0,0 +1,66 @@
...
+                      default="{}")
+    parser.add_option("-p", "--path",
+                      dest="path",
+                      help="The directory where to save the downloaded resource. "
+                           "This path represents the root used when creating the "
+                           "structure. The full path is created by joining this path "
+                           "and the relative_path defined in the resource manifest.")
+    (options, args) = parser.parse_args()
+
+    required_ops = ["resource_type", "resource_name"]
```
gabhijit, line 36: Better to check for `options.resource_type is None and options.resource_name is None`. More Pythonic.

tests/tools/s3_download.py
```diff
@@ -0,0 +1,66 @@
...
+        os.path.join(
+            pathlib.Path(__file__).parent.absolute(),
+            "resource_manifest.json"
+        ))
+
+    res = s3_resource_fetcher.download(
+        resource_type=resource_type,
+        resource_name=resource_name,
+        tags=resource_tags,
+        download_location=location
```
gabhijit, line 63: We can directly use the `option.attribute` here (eg. `options.resource_type`) - So need to get the `options.__dict__` above.

tests/tools/utils.py
```diff
@@ -0,0 +1,99 @@
...
+        specify what tags the resource needs to have, and its version.
+
+        The default version used when downloading is the latest version as defined in the
+        resource manifest."""
+        # TODO: The resource_name does not need to be required, because we can download
+        # resources by type & tags.
+        if version is None:
+            version = self._latest_version
+        if download_location is None:
+            download_location = self._default_download_location
```
gabhijit, line 62: nit: We can simply use `_latest_version` and `_default_download_version` as defaults.

tests/tools/utils.py
```diff
@@ -0,0 +1,99 @@
...
+                self._s3.download_file(self._resource_bucket, object_key, abs_path)
+            downloaded_file_paths.append(abs_path)
+
+        return downloaded_file_paths
+
+    def get_latest_version(self):
+        """Get the latest version as defined in the resource manifest file."""
+        version_re = re.compile("^v[{0-9}]+")
+        versions = [key for key in self._resource_manifest.keys() if version_re.match(key)]
+        return "v{}".format(max(map(lambda v: v[1:], versions)))
```
gabhijit, line 99: Note: This will be lexicographical comparison so "10" < "9", So something like `sorted(versions, key=int, ...)[-1]` or `map(lambda v: int(v[1:])...` is better and since we control the json better to get rid of the `v` in `v1` in the `json` itself?

tests/tools/utils.py
```diff
@@ -0,0 +1,99 @@
...
+            os.makedirs(abs_dir, exist_ok=True)
+            abs_path = os.path.join(abs_dir, resource_name)
+
+            object_key = "{}/{}/{}".format(version, resource_type, resource_name)
+            print("Object to download from S3:", object_key, file=sys.stderr)
+            print("Downloading file to: ", abs_path, file=sys.stderr)
+
+            if not os.path.exists(abs_path):
+                self._s3.download_file(self._resource_bucket, object_key, abs_path)
+            downloaded_file_paths.append(abs_path)
```
gabhijit, line 91: Since we are checking for the existence of the file, should we also have the `sha1sum` inside the `json` and also check that here? Or that's an overkill?

tests/tools/utils.py
andreeaflorescu: I couldn't find a way to do that because they're defined in self. That's why I implemented this workaround.

tests/tools/utils.py
andreeaflorescu: We can check that as well, but I think it should be done as a next step in a subsequent PR because we should also not blindly accept what is already downloaded as the latest version. Instead we should check if any update happened in S3, and in case the local file is older than the one in S3, than download it. We can implement the sha1 in the same PR. Does that make sense?

tests/tools/s3_download.py
andreeaflorescu: If we add more required fields than this check will become rather large. I prefer to have a `required_ops` list.

tests/tools/utils.py
gabhijit: Yes, makes sense. Sorry I missed the `self` part.

tests/tools/s3_download.py
gabhijit: Yes, makes sense. Can we do `getattr(options, op)` instead of `options.__dict__` . Somehow it's best to avoid using the internal `__dict__`  if we can. But that's more like a nit.

tests/tools/utils.py
gabhijit: Yes. I actually added that comment because once I faced an issue due to incorrectly downloaded `kernel.tar.xz` while testing, so definitely a good idea at some point of time but not like super important immediately.

tests/tools/utils.py
gabhijit: A slightly efficient (not that it matters) and somewhat readable way might be - 
`version = version or self.get_latest_version()` and similarly `download_location = download_location or self.get_download_location()`  (and then we do not have to initialize them in the constructor). This will also deal with someone passing "empty" strings accidentally.

src/utils/src/lib.rs
```diff
@@ -0,0 +1,4 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+pub mod resource_download;
```
gabhijit, line 4: Should this be `pub(crate)`?

tests/tools/s3/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+import json
+import os
+import pathlib
+
+from .utils import S3ResourceFetcher
+
+
+def download_resource(resource_type, resource_name, resource_tags="{}", path=None):
```
gabhijit, line 11: Since this is a simple utility function, it does not matter much, but from an API point of view, I  would have `path` as always a required option to this function.; For instance it's really the responsibility of the caller to provide the path to download to (and whether to use a default or a user provided path is what a caller should decide). 

Also, `download_resource` should ideally return a single resource. So that the logic of suitably sorting downloaded resources (and returning the first) remains well within the class itself and the caller just gets one resource. Rather than caller taking out the first resource explicitly. 

And if indeed if multiple are are required to be downloaded (a less common use case), this and the underlying `download` should take `many=False` optional argument. So the caller should specify if they are interested in `many` resources.

tests/tools/s3/utils.py
```diff
@@ -0,0 +1,98 @@
...
+        """Initialize S3 client, manifest of test resources and S3 bucket name."""
+        with open(resource_manifest_path) as json_file:
+            self._resource_manifest = json.load(json_file)
+        self._latest_version = self.get_latest_version()
+        self._resource_bucket = resource_bucket
+        self._s3 = boto3.client(
+            's3',
+            config=botocore.client.Config(signature_version=botocore.UNSIGNED)
+        )
+        if download_location is None:
```
gabhijit, line 39: Once the `download_location` becomes required, this whole thing becomes unnecessary.

tests/tools/s3/utils.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(
+            self,
+            resource_bucket,
+            resource_manifest_path,
+            download_location=None,
+    ):
+        """Initialize S3 client, manifest of test resources and S3 bucket name."""
+        with open(resource_manifest_path) as json_file:
+            self._resource_manifest = json.load(json_file)
+        self._latest_version = self.get_latest_version()
```
gabhijit, line 33: The only place this is used is below inside `download` the `or clause can directly use it, this attribute doesn't become a required one.

tests/test_run_reference_vmm.py
```diff
@@ -1,28 +1,22 @@
 # Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 """Run the reference VMM and shut it down through a command on the serial."""
+from tools.s3 import download_resource
```
gabhijit, line 4: nit: Style wise - it's generally preferred to have imports from standard library first, empty line, then third party external  imports, empty line then imports from within the same package/sub-package - so this should go after `import pytest`

tests/test_run_reference_vmm.py
```diff
@@ -32,6 +26,52 @@ def process_exists(pid):
         return True
 
 
+def default_busybox_bzimage():
```
gabhijit, line 29: See comment on `download_resource` below. (same for all the  `default_*`)

tests/tools/s3_download.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/python3
```
gabhijit, line 1: Just a high level comment - Why not bring everything to this file? The Python code in the tests will use the API provided here and the rust code will use it as a `Command` ?  

This is more of a matter of opinion, but the other two files just have one class and one public function each, which all can be pulled to this file and the class can even be made private(ish) by prefixing it with `_` suggesting users are not advised to initialize this class themselves.

src/utils/src/lib.rs
andreeaflorescu: If this would be pub in crate, then we would not be able to use it outside of `utils`. The `resource_download` is required in external crates like `vmm`.

tests/tools/s3/__init__.py
andreeaflorescu: > Since this is a simple utility function, it does not matter much, but from an API point of view, I would have `path` as always a required option to this function.; For instance it's really the responsibility of the caller to provide the path to download to (and whether to use a default or a user provided path is what a caller should decide).
> 

I am adept of as few parameters as possible, and providing good defaults. In Python this is even easy to achieve. In the normal use case, you should not care where the resources are downloaded because the script already tells you where they are placed, and the location should be good enough for the general scenario (i.e. they're downloaded in `resources` which is the same directory used for the scripts that build said resources). I think only special scenarios require a different location (like for example you want to keep a cache of resources somewhere not in the repository because the repository gets cleaned up after each run). 
 
> Also, `download_resource` should ideally return a single resource. So that the logic of suitably sorting downloaded resources (and returning the first) remains well within the class itself and the caller just gets one resource. Rather than caller taking out the first resource explicitly.
> 

This is not really what I had in mind for this, but then it is also not very well documented. This function is suppose to return all resources that have the specified tags and types (the name should be removed from the required list). That is because I am expecting the test code to evolve such that we parameterize the tests easier than what we do now. For example, let's say that we have support for 2 distros (Ubuntu & Fedora). In the tests I would like to check that all of the available distros are booting under a specified time. Then what I would need to do is the following:
- download_resource(type="kernel", tags="{\"halt-after-boot\": False, \"distro\" = \"*\"} -> ["~/resources/kernel/elf/ubuntu", "~/resources/kernel/elf/fedora", "~/resources/kernel/bzimage/fedora", "~/resources/kernel/elf/ubuntu"]
- for each returned kernel, I want to query all disk resources, that "work-with-kernel":
```
distro_kernels = ..
for kernel in distro_kernel:
     disk = download_resource(type="disk", tags="{\"works-with-kernel\":{}}".format(kernel))
     # add the pair (disk, kernel) to a the list used for parametrizing the test.
```
> And if indeed if multiple are are required to be downloaded (a less common use case), this and the underlying `download` should take `many=False` optional argument. So the caller should specify if they are interested in `many` resources.

I think it is a good idea to have a way of only downloading one. How about doing this reversed though, and have a parameter to just get the first one that matches all tags?

tests/tools/s3/utils.py
andreeaflorescu: I don't think the location should become required. See my previous comment.

tests/tools/s3_download.py
andreeaflorescu: I am not really experienced in Python, so this might be totally wrong :laughing: 

My requirements were:
- reuse the things from utils for both uploading and downloading resources. This means that `S3ResourceUploader` is also defined there and reuses some of the things that are currently available only for `download` (for example getting the latest version)
- the `S3ResourceUploader` is only used from python, so no need to have it as a script.
- the upload functionality is used from both python and rust so we need to have it as a script and as a function

What would be a good way to achieve this?

src/utils/src/lib.rs
gabhijit: Got it! Thanks, The crate here is `utils` I confused it to be `vmm`.

tests/tools/s3/__init__.py
gabhijit: > > Since this is a simple utility function, it does not matter much, but from an API point of view, I would have `path` as always a required option to this function.; For instance it's really the responsibility of the caller to provide the path to download to (and whether to use a default or a user provided path is what a caller should decide).
> 
> I am adept of as few parameters as possible, and providing good defaults. In Python this is even easy to achieve. In the normal use case, you should not care where the resources are downloaded because the script already tells you where they are placed, and the location should be good enough for the general scenario (i.e. they're downloaded in `resources` which is the same directory used for the scripts that build said resources). I think only special scenarios require a different location (like for example you want to keep a cache of resources somewhere not in the repository because the repository gets cleaned up after each run).

Yes, I agree with keeping sensible defaults, so may be it's a default at the outer API level, but the class`s download function (the last one in the call chain) should have it mandatory and download location really need not be a state of the class. In this case it may not matter much because it's the only use of the class 'to download for the tests' or cache.  

The class should simply be a wrapper over the underlying `boto` object corresponding to the `bucket` and `manifest`.  (and can be reused to upload as well, now when I think of it!) 

I also agree, this is more a matter of opinion. :-) 

> 
> > Also, `download_resource` should ideally return a single resource. So that the logic of suitably sorting downloaded resources (and returning the first) remains well within the class itself and the caller just gets one resource. Rather than caller taking out the first resource explicitly.
> 
> This is not really what I had in mind for this, but then it is also not very well documented. This function is suppose to return all resources that have the specified tags and types (the name should be removed from the required list). That is because I am expecting the test code to evolve such that we parameterize the tests easier than what we do now. For example, let's say that we have support for 2 distros (Ubuntu & Fedora). In the tests I would like to check that all of the available distros are booting under a specified time. Then what I would need to do is the following:
> 
>     * download_resource(type="kernel", tags="{"halt-after-boot": False, "distro" = "*"} -> ["~/resources/kernel/elf/ubuntu", "~/resources/kernel/elf/fedora", "~/resources/kernel/bzimage/fedora", "~/resources/kernel/elf/ubuntu"]
> 
>     * for each returned kernel, I want to query all disk resources, that "work-with-kernel":
> 
> 
> ```
> distro_kernels = ..
> for kernel in distro_kernel:
>      disk = download_resource(type="disk", tags="{\"works-with-kernel\":{}}".format(kernel))
>      # add the pair (disk, kernel) to a the list used for parametrizing the test.
> ```
> 

Got it - then may be the function needs some documentation. :-)  and be called `download_resources`

> > And if indeed if multiple are are required to be downloaded (a less common use case), this and the underlying `download` should take `many=False` optional argument. So the caller should specify if they are interested in `many` resources.
> 
> I think it is a good idea to have a way of only downloading one. How about doing this reversed though, and have a parameter to just get the first one that matches all tags?

Yeah then instead of `many=False` (default `first=False`) and caller specifies `first=True` if they want only one.

tests/tools/s3_download.py
gabhijit: The way I would have done is  - 

have only one file `s3_utils.py`  that defines 

1. `S3Wrapper` class and simple convenient APIs like `download_resources` (API if one wants to use it Everything from under `s3/__init__.py` and `s3/utils.py` ) and 
2. Under `if __name__ == '__main__':` define functionality that  can be called as a script. (things in `s3_download.py`) 
Except then the commands need to be explicit like `--download` and `--upload` (with options grouped suitably as needed.)

As a note:  In python when you `import <something>` part from the `if __name__ == '__main__':` does not get executed. So the same file can be used as a lib as well as a script.

src/vmm/src/lib.rs
```diff
@@ -632,6 +645,7 @@ mod tests {
         // Test Case: load a valid elf.
         let mut vmm_config = default_vmm_config();
         vmm_config.kernel_config.path = default_elf_path();
+        println!("{:#?}", vmm_config.kernel_config.path);
```
aghecenco, line 648: Looks like a devtest leftover

tests/tools/s3_download.py
```diff
@@ -28,11 +28,17 @@
                            "This path represents the root used when creating the "
                            "structure. The full path is created by joining this path "
                            "and the relative_path defined in the resource manifest.")
+    parser.add_option("-1",
+                      dest="only_one",
```
aghecenco, line 29: nit: I think `first` is a more idiomatic name

tests/tools/utils.py
```diff
@@ -72,6 +84,11 @@ def download(self, resource_type, resource_name, tags={}, version=None, download
             print("No resources found", file=sys.stderr)
             exit(1)
 
+        if only_one:
+            # When only one resource is needed, we don't need to download all
+            # of them, so we make `resources` a single element array.s
```
aghecenco: nit: trailing `.s`

tests/test_run_reference_vmm.py
```diff
@@ -79,7 +83,7 @@ def start_vmm_process(kernel_path, disk_path=None, num_vcpus=1, mem_size_mib=102
         # The process is still alive.
         pass
 
-    assert(process_exists(vmm_process.pid))
+    assert (process_exists(vmm_process.pid))
```
aghecenco, line 126: nit: are the parentheses needed here?

tests/tools/s3/utils.py
```diff
@@ -121,4 +121,4 @@ def get_latest_version(self):
         """Get the latest version as defined in the resource manifest file."""
         version_re = re.compile("^v[{0-9}]+")
         versions = [key for key in self._resource_manifest.keys() if version_re.match(key)]
-        return "v{}".format(max(map(lambda v: int(v[1:]), versions)))
+        return "v{}".format(max(map(lambda v: int(v[1:]), versions)))
```
aghecenco, line 122: nit: add newline

tests/tools/utils.py
```diff
@@ -0,0 +1,98 @@
...
+            path = os.path.join(
+                os.path.dirname(__file__),
+                "../../resources")
+            self._default_download_location = os.path.abspath(path)
+        else:
+            self._default_download_location = download_location
+
+        print(
+            "Setting default download location: ", self._default_download_location,
+            file=sys.stderr)
```
aghecenco: Instead of `print`ing to `stderr`, you can use the `logging` module which sends `print`s to `stdout` but also has builtin functions that emit logs to `stderr` by default.

```python
log = logging.getLogger(__name__)
log.warning('this goes to stderr')
print('this goes to stdout')
```

tests/tools/utils.py
```diff
@@ -0,0 +1,98 @@
...
+            print("Downloading file to: ", abs_path, file=sys.stderr)
+
+            if not os.path.exists(abs_path):
+                self._s3.download_file(self._resource_bucket, object_key, abs_path)
+            downloaded_file_paths.append(abs_path)
+
+        return downloaded_file_paths
+
+    def get_latest_version(self):
+        """Get the latest version as defined in the resource manifest file."""
```
aghecenco: nit: mix of passive, active, and missing voice in function comments

tests/tools/s3/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        if version in self._resource_manifest.keys():
+            resource_parent = self._resource_manifest[version]
+        else:
+            raise Exception("Invalid version: {}".format(version))
+
+        resources = [r for r in resource_parent
+                     if r["resource_type"] == resource_type
+                     and (resource_name is None or r["resource_name"] == resource_name)
+                     and resource_has_tags(r, tags)]
+        if len(resources) == 0:
```
gabhijit, line 82: Why not raise exception here? Calling `exit(1)` should be responsibility of the caller and not this function.

tests/tools/s3/utils.py
```diff
@@ -0,0 +1,124 @@
...
+            config=botocore.client.Config(signature_version=botocore.UNSIGNED)
+        )
+        if download_location is None:
+            path = os.path.join(
+                os.path.dirname(__file__),
+                "../../../resources")
+            self._default_download_location = os.path.abspath(path)
+        else:
+            self._default_download_location = download_location
+        self.log = logging.getLogger(__name__)
```
gabhijit, line 46: Generally prefer a module level logger, So that all the functions can use it. If indeed you need a class level logger, better to specify `self.__class__.__name__` in this case would evaluate to the `module_name` which is `s3.utils`.

tests/tools/s3/utils.py
```diff
@@ -0,0 +1,124 @@
...
+                     and (resource_name is None or r["resource_name"] == resource_name)
+                     and resource_has_tags(r, tags)]
+        if len(resources) == 0:
+            self.log.error("No resources found")
+            exit(1)
+
+        if first:
+            # When only one resource is needed, we don't need to download all
+            # of them, so we make `resources` a single element array.s
+            resources = resources[:1]
```
gabhijit, line 89: Why not simply return `resources[0]`. If a caller says `first` it's unlikely they are expecting a list.

tests/tools/s3/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+            if not os.path.exists(abs_path):
+                self.log.info("Object to download from S3:", object_key)
+                self.log.info("Downloading file to: ", abs_path)
+                self._s3.download_file(self._resource_bucket, object_key, abs_path)
+            else:
+                print("File already exists locally.", file=sys.stderr)
+            downloaded_file_paths.append(abs_path)
+
+        if len(downloaded_file_paths) == 0:
```
gabhijit, line 111: Same raise Exception here  (May be something like `S3DownloadError`

tests/tools/s3_download.py
```diff
@@ -0,0 +1,52 @@
...
+                      default=False,
+                      action="store_true",
+                      help="Download the first resource that matches the parameters.")
+
+    (options, args) = parser.parse_args()
+
+    if not options.resource_type:
+        parser.error("Missing required parameter: resource_type")
+
+    res = s3_download(
```
gabhijit, line 40: This should be inside a `try` `except` block and **this** should call `sys.exit` depending upon underlying errors.

tests/tools/s3/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+import json
+import os
+import pathlib
+
+from .utils import S3ResourceFetcher
+
+
+def s3_download(resource_type, resource_name, resource_tags="{}", path=None, first=False):
```
gabhijit, line 11: Since this is the real API function, this should add a documentation. Or it should refer to the documentation of underlying `download` function. 

This should definitely add a doc line saying - Any exception raised are not handled, so it's a responsibility of the caller to handle the exception.

tests/test_run_reference_vmm.py
```diff
@@ -33,6 +27,57 @@ def process_exists(pid):
...
+
+def default_disk():
+    return s3_download(
+        resource_type="disk",
+        resource_name="ubuntu-focal-rootfs.ext4",
+        first=True
+    )
+
+
+KERNELS_INITRAMFS = [
```
gabhijit, line 70: Can we add a comment here, when the tests are moved to a class, it's better to perform this inside a `setUp` function.

tests/tools/s3_download.py
```diff
@@ -0,0 +1,52 @@
...
+        options.resource_type,
+        options.resource_name,
+        options.tags,
+        options.path,
+        options.first
+    )
+
+    # When there is a single resource, this is not returned in an array.
+    # Create an array with a single element in this case.
+    res = [res] if options.first is True else res
```
gabhijit, line 50: I am not sure - but will it not become something like `[[/path/to/resource]]`

andreeaflorescu: No, this was previously a bug. Also, if that would be the case, the tests would fail because they expect an array.

tests/test_run_reference_vmm.py
andreeaflorescu: I am not sure I understand what you mean. I don't think these defines will move to a class because they are the parameters of the tests. We were discussing on a separate issue I think about moving the actions you can perform on a `VM` in a class, not the tests.

tests/tools/s3/utils.py
andreeaflorescu: This is needed so that we download exactly one resource from S3. If we don't do this check, then more resources than are needed are going to be downloaded, but then we only return one of them.

tests/tools/s3_download.py
andreeaflorescu: I'll add it.

tests/tools/s3/__init__.py
andreeaflorescu: That sounds like a good idea. I'll add it.

tests/tools/s3_download.py
```diff
@@ -0,0 +1,58 @@
...
+"""Script for downloading the resources required for tests.
+It returns the absolute path of the downloaded resource to stdout."""
+
+import json
+from optparse import OptionParser
+import os
+import pathlib
+
+from utils import S3ResourceFetcher
+
```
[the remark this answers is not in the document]
andreeaflorescu: Done.

tests/tools/s3_download.py
andreeaflorescu: I am going to do this re-arrangement of files in a subsequent PR because it generates lots of merge conflicts.

tests/test_run_reference_vmm.py
gabhijit: I believe `pytest` allow - module level `setUp` and `cleanUp` (I forgot the names). The way it is right now, the functions could get called even when someone `import`s this file. (Not that it matters in this case, because nobody is going to `import` the file.) But a cleaner and readable way is to actually perform actions like resource downloads inside a `setUp` function. 

I saw there was a comment about converting the test cases to a class, I thought it might be a good idea to take that up as a class level `setUp` or perhaps module level `setUp`. So just adding a FIXME or TODO for now is good enough.

tests/tools/s3_download.py
gabhijit: Yes - I agreed along with some other improvements to improvements :-)

tests/tools/s3/utils.py
gabhijit: Oh !!! I am sorry, my bad! I was thinking this was `return`ing resources.

tests/tools/s3_download.py
gabhijit: This was related to my previous comment, Got it!

tests/tools/s3/utils.py
andreeaflorescu: No problem. I'll resolve this.

tests/tools/s3_download.py
andreeaflorescu: It looks like for this we need changes to the logger. I'll open an issue for this instead.

alexandruag (approved):
Looks like these changes will help greatly improve our test framework going forward. Thanks @gabhijit for the detailed review!

aghecenco (approved):

PR #77, opened 2021-02-22
Use debug! macro instead of printlns and asserts

andreeaflorescu:
Defined a helper macro that prints errors only when in debug mode. This is necessary for situations in which we are dealing with potentially malicious input.

The macro is now used in two scenarios:
- for the serial console to print error messages regarding invalid reads/writes
- for handling errors in vcpu exits.

### DIFF
--- a/src/vm-vcpu/src/vcpu/mod.rs
+++ b/src/vm-vcpu/src/vcpu/mod.rs
@@ -19,6 +19,8 @@ use vmm_sys_util::errno::Error as Errno;
 use vmm_sys_util::signal::{register_signal_handler, SIGRTMIN};
 use vmm_sys_util::terminal::Terminal;
 
+use utils::debug;
+
 mod gdt;
 use gdt::*;
 mod interrupts;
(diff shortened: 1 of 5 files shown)


aghecenco (dismissed):

aghecenco (approved):

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -395,7 +397,7 @@ impl KvmVcpu {
                             interrupted_by_signal = true;
                         }
                         _ => {
-                            eprintln!("Emulation error: {}", e);
+                            debug!("Emulation error: {}", e);
```
alexandruag, line 400: Hmm shouldn't we return an error from the function when an emulation error occurs?

andreeaflorescu: I am not really sure. We should probably. I don't know if emulation errors should be considered fatal, in which case we need to exit.

alexandruag: I think we should return an error and relax this later if it turns out too restrictive. No super strong preference, but it sounds like the right thing to IMO.

alexandruag (approved):

PR #78, opened 2021-02-22
serial console: replace asserts with dbg!

andreeaflorescu:
Don't assert on invalid input size when implementing the Bus for the
Serial console. We are dealing with untrusted input, so instead of
asserting, just log an error (in debug mode) when that happens.

Similarly, we should not use expect to convert the PIO offset (which is
an u16), to what the Serial console emulation expects (u8). In case of
invalid length, we can just do nothing.


andreeaflorescu:
Moved the commit from this PR in #77

PR #85, opened 2021-02-25
update dependencies to the latest available version

andreeaflorescu:
Main updates:
- linux-loader to v0.3.0. There is also no need to use a
  patch as we can use the latest version from upstream.
- vm-memory to v0.5.0
- vmm-sys-util to v0.8.0. This comes with interfaces changes.
- kvm-ioctls to v0.7.0.
- event-manager to v0.2.1. This is needed because of the vmm-sys-util
  update.


src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -138,7 +140,8 @@ impl KvmVcpu {
     /// Configure MSRs.
     fn configure_msrs(&self) -> Result<()> {
         let entry_vec = msrs::create_boot_msr_entries();
-        let msrs = Msrs::from_entries(&entry_vec);
+        // This unwrap is safe because we are creating valid kvm_msrs entries.
+        let msrs = Msrs::from_entries(&entry_vec).unwrap();
```
lauralt, line 144: Since we are not checking in any way that `create_boot_msr_entries()` won't create a vec with more than KVM_MAX_MSR_ENTRIES elements, we could stay on the safe side here and propagate the error. Does it sound too exaggerated? :-?

Cargo.lock
```diff
@@ -36,9 +36,9 @@ checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
 
 [[package]]
```
lauralt, line 37: I've noticed that we don't update the `vmm` Cargo.lock. Should we actually remove it?

andreeaflorescu: Yes. That lock is actually ignored, and I think it was added by mistake.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: Right now the function where we get the msrs from is creating a static vector of a known size. If the function `create_boot_msr_entries` ever creates more than 256 entries, I think we would be dealing with a programming error, in which case it is better to immediately exit. I think the guideline is that programming errors should be handled with `panic!`.

This is though a library code, so we can do some extra diligence. How about this course of action instead:
- `create_boot_msr_entries` returns `Msrs` instead of a Vec which is converted to `Msrs` afterwards
- inside the function it's safe to call `Msrs::from_entries(&entry_vec).unwrap()` because the function is the one that creates the `entry_vec` and it knows its exact size

src/vm-vcpu/src/vcpu/msrs.rs
```diff
@@ -13,14 +13,14 @@ use crate::vcpu::msr_index::{
 
 // Creates and populates required MSR entries for booting Linux on X86_64.
```
lauralt, line 14: nit: the description is a bit outdated.

lauralt (dismissed):

src/vm-vcpu/src/vcpu/msrs.rs
```diff
@@ -37,5 +37,8 @@ pub(crate) fn create_boot_msr_entries() -> Vec<kvm_msr_entry> {
             data: u64::from(MSR_IA32_MISC_ENABLE_FAST_STRING),
             ..Default::default()
         },
-    ]
+    ];
+
+    // This unwrap is safe because we are creating valid a fixed size array with msrs entries.
+    Msrs::from_entries(&raw_msrs).unwrap()
```
lauralt, line 43: I would still propagate the error here since we can't be sure `from_entries` won't return in the future other errors as well.

andreeaflorescu: Done.

aghecenco (approved):

lauralt (approved):

Issue #83, opened 2021-02-25
Few issues running tests on dev machine


gabhijit:
Following issues are observed while running tests on my dev machine  (`ubuntu 20.04`/`python 3.8.5`) 

1. When I am trying to run tests on my dev machine,  the test cases wait for a very long time As shown below - 

```shell
$ pytest-3 tests/test_run_reference_vmm.py 
======================================================================== test session starts =========================================================================
platform linux -- Python 3.8.5, pytest-4.6.9, py-1.8.1, pluggy-0.13.1
rootdir: /home/gabhijit/Work/hyphenOs/rust-vmm/vmm-reference
collecting ... 
```
This likely happens due to all the download happening when the `test_run_reference_vmm.py` module gets 'import'ed by `pytest-3`. Likely this can be fixed by moving the `download` code inside a `setUp` function (as discussed in one of the review comments.) 

2. When I run `cargo test --all` Following test case fails 

```shell
running 1 test
test resource_download::tests::test_error_cases ... FAILED

failures:

---- resource_download::tests::test_error_cases stdout ----
thread 'resource_download::tests::test_error_cases' panicked at 'assertion failed: matches!(s3_download("", None) . unwrap_err(), Error :: DownloadError(e) if e
         . contains("Missing required parameter"))', src/utils/src/resource_download.rs:51:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    resource_download::tests::test_error_cases

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '-p utils --lib'
```
This failure happens due to `python3-boto3` not been installed on my dev machine. I believe some updates to documentation should be able to fix this.

3. Documentation does not accurately reflect current state for the testing. Some updates (Just points, actual documentation may be different.) 
- Document that tests require python3 and `pytest` and `boto3` packages installed.
- Give quick instructions to download and install the above packages, both on distro and as a `pip install`
- Any other miscellaneous updates. 

4. Also noticed a small issue - Inside the following file trying to debug one of the above issues - 
```$ cat tests/tools/s3_download.py 
#!/usr/bin/python3
# Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
"""Script for downloading the resources required for tests.
It returns the absolute path of the downloaded resource to stdout."""

[...]

PR #95, opened 2021-03-01
Test improvements to fix some issues in #83

gabhijit:
1. Downloading resources only during running the tests and module import
2. Support for `pyenv` and `virtualenv` python while running tests
3. Updated the documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,13 +117,26 @@ The Python integration tests make use of the VMM in varied configurations that
 aren’t overly complex and illustrate realistic use cases (e.g. one test runs a
 VMM with a virtio block device, one test will run a VMM with PCI, etc.).
 
-To be able to successfully run all the tests in this repo, there are a couple of
-resources that need to be found in `/tmp`. To generate the resources at the
-expected locations, you can run the following command:
+To be able to successfully run all the tests in this repo, pre-created
+resources are stored in S3. The resources are downloaded locally inside the
+`resources` directory. This is handled transparently by the test cases.
+Note: The resources once downloaded are cached, so they are not downloaded on
+every run.
 
-```bash
-sudo ./resources/make_test_resources.sh
+### Running Tests
+
+Recommended way is to run the tests inside a container using the `rustvmm/dev`
+docker image as below. (Note: You may need to run `docker` as `sudo`.)
+
+```shell
+docker run --device=/dev/kvm -it \
+    --security-opt seccomp=unconfined \
+    --volume $(pwd):/vmm-reference rustvmm/dev:v11
+
 ```
+
+Inside the container, to run the tests, first `cd vmm-reference` and then follow
+the instructions as follows.
 
 `vmm-reference` is a
 [workspace](https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html), so to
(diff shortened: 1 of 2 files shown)


tests/test_run_reference_vmm.py
```diff
@@ -67,14 +67,23 @@ def default_disk():
     )
 
 
+@pytest.fixture
```
gabhijit, line 70: A general comment : While these changes are perhaps semantically more correct, I find them a bit kludgy, so if the general opinion is not to take these, we can drop these. A simple `setup_module` approach does not work. This is the least annoying way I could think of : -)

README.md
```diff
@@ -117,13 +117,30 @@ The Python integration tests make use of the VMM in varied configurations that
...
+Please use your distribution specific commands to download this.
+
+For example on Ubuntu based system you can run the following commands to install
+required dependencies -
+`sudo apt-get install python3-boto3 python3-pytest`
+
+or
+
+if you prefer using `pip` to install the dependencies -
+`pip install boto3 pytest`
```
andreeaflorescu, line 141: This is not the recommended way of running the tests. We recommend running the tests with the container which already has all these tools installed. The reason is that tests might fail even if you have all the tools installed if the versions are for example different. Can you please add a sub-section with "Running without the rust-vmm container" and explicitly call out that this is not the recommend way?

tests/test_run_reference_vmm.py
andreeaflorescu: I am not sure these are making the code more readable. I would drop them, and leave the code as it was before. @lauralt what do you think?

README.md
```diff
@@ -117,13 +117,30 @@ The Python integration tests make use of the VMM in varied configurations that
...
+
+### Running Tests
```
andreeaflorescu, line 143: This is not related to this PR, but this section should start with saying that the tests need to be run inside the container:
```bash
docker run --device=/dev/kvm -it \
    --security-opt seccomp=unconfined \
    --volume $(pwd):/vmm-reference rustvmm/dev:v11
# in the container
pytest rust-vmm-ci tests/test_run_reference_vmm.py
```

gabhijit: Are you sure `pytest rust-vmm-ci tests/test_run_reference_vmm.py`. I tried to quickly spawn the container and try to run the command that didn't work. Following works though - 
```shell
cd vmm-reference;  pytest tests/test_run_reference_vmm.py
```

andreeaflorescu: Right! You need to first go to the `vmm-reference` directory.

README.md
```diff
@@ -117,14 +117,47 @@ The Python integration tests make use of the VMM in varied configurations that
...
-To be able to successfully run all the tests in this repo, there are a couple of
-resources that need to be found in `/tmp`. To generate the resources at the
-expected locations, you can run the following command:
+To be able to successfully run all the tests in this repo, pre-created
+resources are stored in S3. The resources are downloaded locally inside the
+`resources` directory. This is handled transparently by the test cases.
+Note: The resources once downloaded are cached, so they are not downloaded on
+every run.
+
+### Dependencies for Testing
```
andreeaflorescu, line 126: Can we move this section after the `Running Tests` section?

gabhijit: I am just thinking whether this section is needed at all :-) Because the container image thing works just fine. If people indeed want to work outside the container, let the figure out.  It's more like an advanced usage, likely not required in the `README.md` file.

tests/test_run_reference_vmm.py
lauralt: So the "only" advantage with this approach is saving time when you first run the tests (and when you run them again there's no time difference between this solution and the way it was before)? If yes, I would not add these fixtures.

Otherwise, PR looks good to me. Thanks for the contribution!

gabhijit: I'll drop the commit then!

lauralt (approved):

andreeaflorescu (approved):

PR #111, opened 2021-03-22
Add piping guest STDIN/STDOUT to named pipes support

borna-blazevic:
This PR adds support for piping guest STDIN/STDOUT to named pipes so it can be consumed there. When named pipes are enabled the starting process of the guest VM is:

1. The VMM creates a FIFO (Named pipe) for its input.
2. The VMM opens the FIFO for reading, which blocks until a producer opens the FIFO for writing.
3. The VMM creates a FIFO (Named pipe) for its output.
4. If the guest VM has anything to write the VMM opens the FIFO for writing, writes to it and closes it afterwards.

Also, tests for this scenario has been added to integration tests - `tests/test_run_reference_vmm.py::test_serial_console_piping` where this process is shown.

### DIFF
--- a/src/utils/src/lib.rs
+++ b/src/utils/src/lib.rs
@@ -1,8 +1,13 @@
 // Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
+use std::ffi::CString;
+use std::path::Path;
+use std::{io, result};
+
 pub mod resource_download;
 
+// -----------------------------------------------------------------------------
 /// A macro for printing errors only in debug mode.
 #[macro_export]
 #[cfg(debug_assertions)]
@@ -17,4 +22,37 @@ macro_rules! debug {
     ($( $args:expr ),*) => {
         ()
     };
+}
+
+// -----------------------------------------------------------------------------
+/// Enum containing error types.
+pub enum FifoError {
+    Io(io::Error),
+    Path,
+}
+
+impl std::fmt::Display for FifoError {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "Erorr: {}", (*self))
+    }
+}
+
+/// A safe wrapper around libc::mkfifo().
+pub fn mkfifo<P: AsRef<Path>>(fifo_path: P, mode: u32) -> result::Result<(), FifoError> {
+    let fifo_path_str = fifo_path.as_ref().to_str().ok_or(FifoError::Path)?;
+    match CString::new(fifo_path_str) {
+        Ok(fifo_path_cstr) => {
+            // fifo_path_cstr was checked to be valid and the return of the unsafe code was checked
+            let ret = unsafe { libc::mkfifo(fifo_path_cstr.as_ptr(), mode) };
+            if ret != 0 {
+                Err(FifoError::Io(io::Error::last_os_error()))
+            } else {
+                Ok(())
+            }
+        }
+        Err(_) => Err(FifoError::Io(io::Error::new(
+            io::ErrorKind::InvalidInput,
+            "pipe name cannot contain null-bytes",
+        ))),
+    }
 }
--- a/src/vmm/src/config/mod.rs
+++ b/src/vmm/src/config/mod.rs
@@ -32,6 +32,8 @@ pub enum ConversionError {
     ParseNet(String),
     /// Failed to parse the string representation for the block.
     ParseBlock(String),
+    /// Failed to parse the string representation for the serial config.
+    ParseSerial(String),
 }
 
 impl ConversionError {
@@ -49,6 +51,9 @@ impl ConversionError {
     }
     fn new_net<T: fmt::Display>(err: T) -> Self {
         Self::ParseNet(err.to_string())
+    }
+    fn new_serial<T: fmt::Display>(err: T) -> Self {
+        Self::ParseSerial(err.to_string())
     }
 }
 
@@ -68,6 +73,7 @@ impl fmt::Display for ConversionError {
             ParseVcpus(ref s) => write!(f, "Invalid input for vCPUs: {}", s),
             ParseNet(ref s) => write!(f, "Invalid input for network: {}", s),
             ParseBlock(ref s) => write!(f, "Invalid input for block: {}", s),
+            ParseSerial(ref s) => write!(f, "Invalid input for serial: {}", s),
         }
     }
 }
@@ -232,6 +238,50 @@ impl TryFrom<&str> for NetConfig {
     }
 }
 
+/// Serial communication configuration
+#[derive(Clone, Debug, PartialEq)]
+pub struct SerialConfig {
+    /// Optional path to pipe the input of the guest OS.
+    pub serial_input: Option<PathBuf>,
+    /// Optional path to pipe the output of the guest OS.
+    pub serial_output: Option<PathBuf>,
+}
+
+impl Default for SerialConfig {
+    fn default() -> Self {
+        SerialConfig {
+            serial_input: None,
+            serial_output: None,
+        }
+    }
+}
+
+impl TryFrom<&str> for SerialConfig {
+    type Error = ConversionError;
+
+    fn try_from(serial_console_str: &str) -> Result<Self, Self::Error> {
+        let mut arg_parser = CfgArgParser::new(serial_console_str);
+
+        let serial_input = arg_parser
+            .value_of::<String>("serial_input")
+            .map_err(ConversionError::new_serial)?
+            .map(PathBuf::from);
+
+        let serial_output = arg_parser
+            .value_of::<String>("serial_output")
+            .map_err(ConversionError::new_serial)?
+            .map(PathBuf::from);
+
+        arg_parser
+            .all_consumed()
+            .map_err(ConversionError::new_serial)?;
+        Ok(SerialConfig {
+            serial_input,
+            serial_output,
+        })
+    }
+}
+
 /// Block device configuration
 #[derive(Clone, Debug, PartialEq)]
 pub struct BlockConfig {
@@ -271,6 +321,8 @@ pub struct VMMConfig {
     pub net_config: Option<NetConfig>,
     /// Block device configuration.
     pub block_config: Option<BlockConfig>,
+    /// Serial communication configuration.
+    pub serial_config: SerialConfig,
 }
 
 #[cfg(test)]
@@ -385,6 +437,67 @@ mod tests {
         // Test case: unused parameters
         let block_str = "path=/foo/bar,blah=blah";
         assert!(BlockConfig::try_from(block_str).is_err());
+    }
+
+    #[test]
+    fn test_serial_config() {
+        // Testing the positive case with both input and output
+        let serial_str = r#"serial_input=/foo/bar/in,serial_output=/foo/bar/out"#;
+        let serial_cfg = SerialConfig::try_from(serial_str).unwrap();
+        let expected_cfg = SerialConfig {
+            serial_input: Some(PathBuf::from(String::from("/foo/bar/in"))),
+            serial_output: Some(PathBuf::from(String::from("/foo/bar/out"))),
+        };
+        assert_eq!(serial_cfg, expected_cfg);
+
+        // Testing that the config is succesfully created with empty str for input
+        let serial_str_empty_in = r#"serial_input=,serial_output=/foo/bar/out"#;
+        let serial_cfg_empty_in = SerialConfig::try_from(serial_str_empty_in).unwrap();
+        let expected_cfg_empty_in = SerialConfig {
+            serial_input: None,
+            serial_output: Some(PathBuf::from(String::from("/foo/bar/out"))),
+        };
+        assert_eq!(serial_cfg_empty_in, expected_cfg_empty_in);
+
+        // Testing that the config is succesfully created with empty str for output
+        let serial_str_empty_out = r#"serial_input=/foo/bar/in,serial_output="#;
+        let serial_cfg_empty_out = SerialConfig::try_from(serial_str_empty_out).unwrap();
+        let expected_cfg_empty_out = SerialConfig {
+            serial_input: Some(PathBuf::from(String::from("/foo/bar/in"))),
+            serial_output: None,
+        };
+        assert_eq!(serial_cfg_empty_out, expected_cfg_empty_out);
+
+        // Testing that the config is succesfully created without input
+        let serial_str_no_in = r#"serial_output=/foo/bar/out"#;
+        let serial_cfg_no_in = SerialConfig::try_from(serial_str_no_in).unwrap();
+        let expected_cfg_no_in = SerialConfig {
+            serial_input: None,
+            serial_output: Some(PathBuf::from(String::from("/foo/bar/out"))),
+        };
+        assert_eq!(serial_cfg_no_in, expected_cfg_no_in);
+
+        // Testing that the config is succesfully created without output
+        let serial_str_no_out = r#"serial_input=/foo/bar/in"#;
+        let serial_cfg_no_out = SerialConfig::try_from(serial_str_no_out).unwrap();
+        let expected_cfg_no_out = SerialConfig {
+            serial_input: Some(PathBuf::from(String::from("/foo/bar/in"))),
+            serial_output: None,
+        };
+        assert_eq!(serial_cfg_no_out, expected_cfg_no_out);
+
+        // Testing that the config is succesfully created for an empty string
+        let serial_str_empty = r#""#;
+        let serial_cfg_empty = SerialConfig::try_from(serial_str_empty).unwrap();
+        let expected_cfg_empty = SerialConfig {
+            serial_input: None,
+            serial_output: None,
+        };
+        assert_eq!(serial_cfg_empty, expected_cfg_empty);
+
+        // Testing that the invalid string raises an error
+        let serial_str_invalid = r#"invalid"#;
+        assert!(SerialConfig::try_from(serial_str_invalid).is_err());
     }
 
     #[test]
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -7,7 +7,7 @@ use std::convert::TryFrom;
 #[cfg(target_arch = "aarch64")]
 use std::convert::TryInto;
 use std::fs::File;
-use std::io::{self, stdin, stdout};
+use std::io::{self, stdin, Write};
 use std::ops::DerefMut;
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicBool, Ordering};
@@ -59,7 +59,7 @@ use devices::virtio::block::{self, BlockArgs};
 use devices::virtio::net::{self, NetArgs};
 use devices::virtio::{Env, MmioConfig};
 
-use devices::legacy::{EventFdTrigger, SerialWrapper};
+use devices::legacy::{ConsoleReaderWrapper, ConsoleWriterWrapper, EventFdTrigger, SerialWrapper};
 use vm_vcpu::vcpu::VcpuState;
 use vm_vcpu::vm::{self, ExitHandler, KvmVm, VmState};
 #[cfg(target_arch = "x86_64")]
@@ -187,6 +187,7 @@ pub struct Vmm {
     exit_handler: WrappedExitHandler,
     block_devices: Vec<Arc<Mutex<Block>>>,
     net_devices: Vec<Arc<Mutex<Net>>>,
+    serial_config: SerialConfig,
     // TODO: fetch the vcpu number from the `vm` object.
     // TODO-continued: this is needed to make the arm POC work as we need to create the FDT
     // TODO-continued: after the other resources are created.
@@ -283,6 +284,7 @@ impl TryFrom<VMMConfig> for Vmm {
             exit_handler: wrapped_exit_handler,
             block_devices: Vec::new(),
             net_devices: Vec::new(),
+            serial_config: config.serial_config,
             #[cfg(target_arch = "aarch64")]
             num_vcpus: config.vcpu_config.num as u64,
         };
@@ -441,11 +443,17 @@ impl Vmm {
     // Create and add a serial console to the VMM.
     fn add_serial_console(&mut self) -> Result<()> {
         // Create the serial console.
+        let serial;
         let interrupt_evt = EventFdTrigger::new(libc::EFD_NONBLOCK).map_err(Error::IO)?;
-        let serial = Arc::new(Mutex::new(SerialWrapper(Serial::new(
-            interrupt_evt.try_clone().map_err(Error::IO)?,
-            stdout(),
-        ))));
+
+        let input = ConsoleReaderWrapper::new(self.serial_config.serial_input.clone());
+        let output = ConsoleWriterWrapper::new(self.serial_config.serial_output.clone());
+
+        let out_writer: Box<dyn Write + Send> = Box::new(output);
+        serial = Arc::new(Mutex::new(SerialWrapper {
+            serial: Serial::new(interrupt_evt.try_clone().map_err(Error::IO)?, out_writer),
+            console_reader: input,
+        }));
 
         // Register its interrupt fd with KVM. IRQ line 4 is typically used for serial port 1.
         // See more IRQ assignments & info: https://tldp.org/HOWTO/Serial-HOWTO-8.html
@@ -730,6 +738,7 @@ mod tests {
             vcpu_config: VcpuConfig { num: NUM_VCPUS },
             block_config: None,
             net_config: None,
+            serial_config: SerialConfig::default(),
         }
     }
 
@@ -763,6 +772,7 @@ mod tests {
             exit_handler,
             block_devices: Vec::new(),
             net_devices: Vec::new(),
+            serial_config: vmm_config.serial_config,
             #[cfg(target_arch = "aarch64")]
             num_vcpus: vmm_config.vcpu_config.num as u64,
         }
--- a/tests/test_run_reference_vmm.py
+++ b/tests/test_run_reference_vmm.py
@@ -86,7 +86,7 @@ and run locally.
 """
 
 
-def start_vmm_process(kernel_path, disk_path=None, num_vcpus=1, mem_size_mib=1024):
+def start_vmm_process(kernel_path, disk_path=None, num_vcpus=1, mem_size_mib=1024, serial_in="", serial_out=""):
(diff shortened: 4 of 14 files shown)


sameo:
Out of curiosity, why not passing a unix socket? Or even support both?

borna-blazevic:
> Out of curiosity, why not passing a unix socket? Or even support both?

Thank You for the input @sameo, Unix socket would basically be implemented in the same way, but it is a bit more complicated on the consumer side. I can see it also be useful, but maybe in another PR?

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,69 @@
...
+
+#![cfg(target_arch = "x86_64")]
+
+use std::io::{Read, Write};
+
+use std::fs::{File};
+use std::os::unix::io::{AsRawFd, RawFd};
+
+use nix::sys::stat;
+use nix::unistd;
```
aghecenco, line 12: I think `nix` is a pretty big (new) dependency to pull in just for named pipes. `libc` is already a dependency and a basic safe wrapper over `libc::mkfifo` could be added to the `utils` crate:
```rust
pub enum Error {
    Io(io::Error),
    Path,
}

pub fn mkfifo<P: AsRef<Path>>(fifo_path: P, mode: u32) -> result::Result<(), Error> {
    let fifo_path_str = fifo_path.as_ref().to_str().ok_or(Error::Path)?;
    let ret = unsafe { libc::mkfifo(fifo_path_str.as_ptr() as *const i8, mode) };
    if ret != 0 {
        Err(Error::Io(io::Error::last_os_error()))
    } else {
        Ok(())
    }
}
```
(This is just a suggestion and can be simplified a lot by just working with a `&str` instead of generics.)

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,69 @@
...
+/// Newtype for piping the output of guest cmd.
+pub(crate) struct PipeReaderWrapper(PathBuf, File);
+
+impl PipeReaderWrapper {
+    pub fn new(buf: PathBuf) -> PipeReaderWrapper {
+        match unistd::mkfifo(&buf, stat::Mode::S_IRWXO) {
+            Ok(_) => (),
+            Err(err) => panic!("Error creating fifo: {}", err),
+        }
+        let pipe = match File::open(&buf) {
```
aghecenco, line 24: I'm thinking of this a bit differently: what if you created the `Read`er somewhere else, and worked with that instead, passing it on to different implementors? That way you could have a generic serial reader that only cares about the sink being `Read` and `AsRawFd`, and 2+ different instantiations:
1. `io::Stdin`, which conveniently is both `Read` and `AsRawFd` already
2. a pipe
3. (why not?) the unix domain socket suggested by @sameo 

This is an idea somewhat orthogonal to the changes in this PR - the rest of my comments directly address the individual changes, but I'd like to get your opinion on this approach.

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,69 @@
...
+            Ok(_) => (),
+            Err(err) => panic!("Error creating fifo: {}", err),
+        }
+        return PipeWriterWrapper(buf);
+    }
+}
+
+impl Write for PipeWriterWrapper {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let mut pipe = match File::create(&self.0) {
```
aghecenco, line 58: I think this needs `OpenOptions` to make sure it's writable and to control the opening behavior (`O_CREAT`, `O_TRUNC` etc).

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,69 @@
...
+            Err(err) => panic!("couldn't open pipe: {}", err),
```
aghecenco, line 59: I don't think the code should panic here. It's just an error that can be propagated further up.

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,69 @@
+// Copyright 2020 Sartura All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+#![cfg(target_arch = "x86_64")]
```
aghecenco, line 4: I don't think this is needed, the top-level crate configuration is already set to `x86_64`-exclusive.

src/vmm/src/config/mod.rs
```diff
@@ -140,6 +140,8 @@ pub struct KernelConfig {
     pub path: PathBuf,
     /// Start address for high memory.
     pub himem_start: u64,
+    /// Optional path to pipe the input and output of the gues OS
+    pub serial_console_path: Option<String>,
```
aghecenco, line 144: I don't think this is conceptually a guest kernel configuration. Since it's an option for how to operate one of the VMM's devices, I'd rather see it in a new `SerialConfig` (or something...I'm bad with names) struct.

src/vmm/src/lib.rs
```diff
@@ -379,11 +381,33 @@ impl VMM {
...
-        let serial = Arc::new(Mutex::new(SerialWrapper(Serial::new(
-            interrupt_evt.try_clone().map_err(Error::IO)?,
-            stdout(),
-        ))));
+        match &self.kernel_cfg.serial_console_path {
+            Some(x) => {
+                let input_pipe_string = x.to_owned() + "_in";
+                let output_pipe_string = x.to_owned() + "_out";
+                let input_pipe_path = PathBuf::from(input_pipe_string);
+                let output_pipe_path = PathBuf::from(output_pipe_string);
```
aghecenco, line 391: This looks like it would be easier if we just let the user specify both paths, instead of creating files starting from a prefix, which some users could find confusing. I also think it'd be easier to work with `&str`s or `P: AsRef<Path>` instead of owned `String`s and `PathBuf`s to make use of borrowing wherever possible.

src/vmm/src/lib.rs
```diff
@@ -379,11 +381,33 @@ impl VMM {
...
+
+                let input = PipeReaderWrapper::new(input_pipe_path);
+                let output = PipeWriterWrapper::new(output_pipe_path);
+
+                let out_writer: Box<dyn Write + Send> = Box::new(output);
+                serial = Arc::new(Mutex::new(SerialWrapper(
+                    Serial::new(interrupt_evt.try_clone().map_err(Error::IO)?, out_writer),
+                    Some(input),
+                )));
+                println!("Finished with pipe setup!");
```
aghecenco, line 401: Is this a leftover debug print?

src/vmm/src/serial.rs
```diff
@@ -14,42 +14,78 @@ use vm_device::{
...
             }
-        }
+        };
     }
 
     fn init(&mut self, ops: &mut EventOps) {
         // Hook to stdin events.
-        ops.add(Events::new(&stdin(), EventSet::IN))
-            .expect("Failed to register serial input event");
+        match &self.1 {
```
aghecenco, line 79: This would also be collapsed into a single arm if the `Read`er part was generic, I believe.

andreeaflorescu:
@borna-blazevic do you need help in addressing the remaining comments on this PR?

borna-blazevic:
Hi Andrea, sorry for the small delay, I think the changeset is now in a shape where you can take another look. All the comments should be addressed. The only thing I am struggling with a bit is the tests - as you can see the integration tests fail, but I am not able to recreate the same behaviour locally (tried it on a clean PC as well just as a sanity check and all the tests pass). Do you have any input regarding this?

tests/test_run_reference_vmm.py
```diff
@@ -139,6 +144,17 @@ def shutdown(vmm_process):
...
 
+def shutdown_piped(vmm_process, vmm_input_fd):
+    os.write(vmm_input_fd, b'reboot -f\n')
+
+    # If the process hasn't ended within 3 seconds, this will raise a
+    # TimeoutExpired exception and fail the test.
+    vmm_process.wait(timeout=3)
+
+    os.remove(NAMED_PIPES["in"])
+    os.remove(NAMED_PIPES["out"])
```
andreeaflorescu, line 155: We should remove them when the test fails as well. I think a better approach would be to not have them as global variables, and just define them in the test. Then, in the test we can initialize them to temporary files. It looks like this can be achieved with https://docs.pytest.org/en/6.2.x/tmpdir.html, but I haven't looked in depth at it.

src/utils/src/lib.rs
```diff
@@ -1,8 +1,16 @@
 // Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+#[macro_use]
+extern crate enum_display_derive;
```
andreeaflorescu, line 4: We are trying to keep the external dependencies to a minimum, can we remove this one?

src/vmm/src/config/mod.rs
```diff
@@ -211,6 +217,54 @@ impl TryFrom<&str> for NetConfig {
...
+    }
+}
+
+impl TryFrom<&str> for SerialConfig {
+    type Error = ConversionError;
+
+    fn try_from(serial_console_str: &str) -> Result<Self, Self::Error> {
+        let mut arg_parser = CfgArgParser::new(serial_console_str);
+
+        let serial_input: Option<String> = arg_parser
```
andreeaflorescu, line 244: We don't need to explicitly mention the type here.

src/vmm/src/config/mod.rs
```diff
@@ -211,6 +217,54 @@ impl TryFrom<&str> for NetConfig {
...
+            .value_of("serial_input")
+            .map_err(ConversionError::new_serial)?;
```
andreeaflorescu, line 246: We can chain an `and_then` here to simplify the code that initializes the `SerialConfig`. If we initialize this directly to an `Option<PathBuf>`, then we will not need the match for building the `SerialConfig`.

We can have it as:
```rust
        let serial_input = arg_parser
            .value_of("serial_input")
            .map_err(ConversionError::new_serial)?
            .and_then(|inner: String| Some(PathBuf::from(inner)));
```

src/vmm/src/config/mod.rs
```diff
@@ -362,6 +418,17 @@ mod tests {
         assert!(BlockConfig::try_from(block_str).is_err());
     }
 
+    #[test]
+    fn test_serial_config() {
```
andreeaflorescu, line 422: Should we add some tests with only the input set and not the output, and the output set, but not the input? A test case for an invalid setup would also be nice.

andreeaflorescu (commented):
I did a quick look, I still need to look more in depth.

I also ran the tests locally and they pass. I'll try to run them in a similar environment as the CI to understand why they fail there.

borna-blazevic:
I tried to address all of your comments with the latest changes, and the CI is green now (not sure if you made some changes there or was it the result of the latest changes) so when you have time you could take another look.

lperkov:
Hello @andreeaflorescu, just curious if you could provide feedback on this pull request.

Thanks!

andreeaflorescu:
Hey @borna-blazevic and @lperkov. I missed the notifications on this PR, sorry about that. I'll take a look in the next couple of days.

src/utils/src/lib.rs
```diff
@@ -18,3 +23,30 @@ macro_rules! debug {
         ()
     };
 }
+
+// -----------------------------------------------------------------------------
+/// Enum containing error types.
+pub enum Error {
```
andreeaflorescu, line 29: I find this name to be too generic for having it in the lib.rs of the utils module. I would propose to rename it `FifoError`.

src/utils/src/lib.rs
```diff
@@ -18,3 +23,30 @@ macro_rules! debug {
...
+        write!(f, "Erorr: {}", (*self))
+    }
+}
+
+/// A safe wrapper around libc::mkfifo().
+pub fn mkfifo<P: AsRef<Path>>(fifo_path: P, mode: u32) -> result::Result<(), Error> {
+    let fifo_path_str = fifo_path.as_ref().to_str().ok_or(Error::Path)?;
+    let fifo_path_cstr = CString::new(fifo_path_str).unwrap();
+
+    let ret = unsafe { libc::mkfifo(fifo_path_cstr.as_ptr(), mode) };
```
andreeaflorescu, line 45: Unsafe code blocks should be accompanied by comments saying why they're actually safe. In this case we can say that we checked that `fifo_path_cstr` is valid, and we checked the return of the unsafe code.

src/utils/src/lib.rs
```diff
@@ -18,3 +23,30 @@ macro_rules! debug {
...
+impl std::fmt::Display for Error {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "Erorr: {}", (*self))
+    }
+}
+
+/// A safe wrapper around libc::mkfifo().
+pub fn mkfifo<P: AsRef<Path>>(fifo_path: P, mode: u32) -> result::Result<(), Error> {
+    let fifo_path_str = fifo_path.as_ref().to_str().ok_or(Error::Path)?;
+    let fifo_path_cstr = CString::new(fifo_path_str).unwrap();
```
andreeaflorescu, line 43: Why is this `unwrap` safe? Looking at the documentation of `Cstring::new`, it fails when the string contains 0 bytes in it:
https://doc.rust-lang.org/std/ffi/struct.CString.html#method.new

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,127 @@
...
+
+/// A wrapper structure around that provides basic read functionality backed by
+/// a named pipe implementation.
+pub(crate) struct PipeReaderWrapper(PathBuf, File);
+
+impl PipeReaderWrapper {
+    pub fn new(buf: PathBuf) -> PipeReaderWrapper {
+        match utils::mkfifo(&buf, libc::S_IRWXO) {
+            Ok(_) => (),
+            Err(err) => panic!("Error creating fifo: {}", err),
```
andreeaflorescu, line 17: We should return an error instead of panic. Same for all the panics in this file.

Once we define an error, we should also use `map_err` and convert these lines to 1-liners.

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+        // let pipe = match File::open(&buf) {
```
andreeaflorescu, line 19: nit: leftover comment?

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,127 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use std::fs::{File, OpenOptions};
+use std::io::{Read, Write};
+use std::os::unix::io::{AsRawFd, RawFd};
+use std::path::PathBuf;
+
+/// A wrapper structure around that provides basic read functionality backed by
+/// a named pipe implementation.
+pub(crate) struct PipeReaderWrapper(PathBuf, File);
```
andreeaflorescu, line 11: Why not create a structure with named fields instead? It makes it easier to follow the code I think.

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,127 @@
...
+        let mut pipe = match File::create(&self.0) {
+            Err(err) => {
+                println!("couldn't open pipe: {}", err);
+                return Err(err);
+            }
+            Ok(input) => input,
+        };
+        pipe.write(buf)
+    }
+    fn flush(&mut self) -> std::io::Result<()> {
```
andreeaflorescu, line 66: nit: white space needed before function definition.

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,127 @@
...
+            Ok(_) => (),
+            Err(err) => panic!("Error creating fifo: {}", err),
+        }
+        PipeWriterWrapper(buf)
+    }
+}
+
+impl Write for PipeWriterWrapper {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        let mut pipe = match File::create(&self.0) {
```
andreeaflorescu, line 57: I don't understand this part. Why do we need to call `File::create` here? Isn't the file already created when calling new? Shouldn't we be able to just call `open` on the `PathBuf`?

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,127 @@
+// Copyright 2020 Sartura All Rights Reserved.
```
andreeaflorescu, line 1: What is the reason for having these wrappers here instead of in the `utils`?

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,127 @@
...
+            drop(output);
+        });
+        let mut read_buffer: [u8; 1] = [0];
+        let input_pipe_path = PathBuf::from("test_in");
+        let mut input = PipeReaderWrapper::new(input_pipe_path);
+        let res = input.read(&mut read_buffer);
+        assert!(res.is_ok() && !res.is_err());
+        assert_eq!(read_buffer[0], 1);
+        drop(input);
+        let res = remove_file("test_in");
```
andreeaflorescu, line 99: If any checks before this call fail, then we will have `test_in` on the host even after the test finishes.
To avoid such situations, we are using `tempfile`. You can check an example here: https://github.com/rust-vmm/vmm-reference/blob/30d7adf450be68ca95213e1bd779764ad2b6f1fc/src/devices/src/virtio/block/mod.rs#L124

src/vmm/src/pipe.rs
```diff
@@ -0,0 +1,127 @@
...
+            };
+            let res = output.write(&write_buffer);
+            assert!(res.is_ok() && !res.is_err());
+            drop(output);
+        });
+        let mut read_buffer: [u8; 1] = [0];
+        let input_pipe_path = PathBuf::from("test_in");
+        let mut input = PipeReaderWrapper::new(input_pipe_path);
+        let res = input.read(&mut read_buffer);
+        assert!(res.is_ok() && !res.is_err());
```
andreeaflorescu, line 96: `res.is_ok` and `res.is_err` are mutually exclusive. There is no reason to check for both, checking `is_ok` will suffice.

src/vmm/src/serial.rs
```diff
@@ -3,31 +3,32 @@
...
     MutDevicePio,
 };
 use vm_superio::Serial;
 use vmm_sys_util::epoll::EventSet;
 
-use utils::debug;
-
 /// Newtype for implementing `event-manager` functionalities.
-pub(crate) struct SerialWrapper<W: Write>(pub Serial<W>);
+pub(crate) struct SerialWrapper<W: Write>(pub Serial<W>, pub ConsoleReaderWrapper);
```
andreeaflorescu, line 19: This is making the code hard to read. We've been using the newtype pattern only for when we have one type as inner type. Since we are now having 2 types here, we should define it as a struct with named fields.

tests/test_run_reference_vmm.py
```diff
@@ -101,7 +100,8 @@ def start_vmm_process(kernel_path, disk_path=None, num_vcpus=1, mem_size_mib=102
         "--kernel", "cmdline={},path={},himem_start={}".format(
             cmdline, kernel_path, himem_start
         ),
-        "--vcpu", "num={}".format(num_vcpus)
+        "--vcpu", "num={}".format(num_vcpus),
+        "--serial", "serial_input={},serial_output={}".format(serial_in, serial_out)
```
andreeaflorescu, line 104: I think we shouldn't forward the `serial_input` and `serial_output` to the vmm process if they're the empty string. This will help with keeping the default behavior as we had it before.

tests/test_run_reference_vmm.py
```diff
@@ -139,6 +139,14 @@ def shutdown(vmm_process):
     vmm_process.wait(timeout=3)
 
 
+def shutdown_piped(vmm_process, vmm_input_fd):
```
andreeaflorescu, line 142: At this point I don't think it's feasible to continue to use `vmm_process` as we had before. It becomes more apparent that we need to define a `Vmm` class that keeps the `stdin` and `stdout` configuration, along with the other configurations that are needed in order to not duplicate the `shutdown` and the `expect_string` functions.

At a minimum, if we do not implement this in this PR, we should have a generic shutdown function that receives as a parameter the file descriptor, and then write a variant for it where the fd is actually stdin. Same approach should be used for `expect_string`

tests/test_run_reference_vmm.py
```diff
@@ -370,3 +406,32 @@ def test_reference_vmm_mem(kernel):
         expect_mem(vmm_process, expected_mem_mib)
 
         shutdown(vmm_process)
+
+
+@pytest.mark.parametrize("kernel", KERNELS_INITRAMFS)
+def test_serial_console_piping(kernel, tmp_path):
```
andreeaflorescu, line 412: What is `tmp_path`? Is it some sort of keyword that gives you a temporary directory?

andreeaflorescu:
@borna-blazevic the test that is running the VMM apparently hanged. We added a change to the test so that it fails after 1 minute. Can you rebase your PR and also check if the test is failing locally as well?

src/vmm/src/pipe.rs
borna-blazevic: ```File::create``` opens a file in write-only mode, whereas ```File::open``` opens a file in read-only mode. Since this is the writer this seems fine. I agree that it's weirdly named, but that's the way it's [documented](https://doc.rust-lang.org/std/fs/struct.File.html#method.create).

src/vmm/src/pipe.rs
borna-blazevic: It seemed appropriate since it's related to the ```serial.rs``` file if You disagree we can change it?

borna-blazevic:
@andreeaflorescu I have addressed all of the open points and rebased everything on top of the newest changes so could you please take a look when you have time?

Thanks

andreeaflorescu:
> Can we achieve the same by using stdin/out redirection for the VMM process such that input and/or output come from named pipes (for example)? If that's an option, does it happen to work with the basic console support which existed before this PR?

@borna-blazevic can you please take a look at this question? This is something that we missed in the beginning, and we were wondering if you also explored this option.

borna-blazevic:
Sorry for the delay, I was otherwise occupied and had to experiment a bit with this before I could write a proper answer. I have tried to run a VMM process with in/out redirections to named pipes and I observed the following:
- When redirecting the output, the behaviour is as expected (as long as there is a process on the other side to read from the FIFO everything runs smoothly)
- When redirecting input, things get a bit weird -> the VMM process hangs initially before the "other side" puts something in the FIFO, at which point it successfully boots a VM, but subsequent writes to the input FIFO are ignored (I verified that the commands were in fact written to the input FIFO by reading it with cat after VMM was shutdown).

I tried this both in the shell and by using a bit "fancier" python wrappers (namely the subprocess module) which, other than actual files, allows for virtual FIFOs and the behaviour is more or less the same (the python communicate command hangs forever which would indicate that the input pipe is never read from by the VMM process).

But other than this, I would like to point out that the additional level of abstraction that we introduce here is not only relevant for named pipes, but allows for further extensions with less effort (e.g. TCP socket, Unix socket) which we already have in the pipeline and would happily merge as a follow-up.

borna-blazevic:
Hi, @andreeaflorescu, @alexandruag just a small reminder, if you could provide me with some feedback regarding our discussion, and whether the pr is now ok to be merged or if there is further things to adapt?
Thanks :)

andreeaflorescu:
Hey @borna-blazevic, I was on PTO. I will be back at this PR early next week, sorry for the delay.

How are you configuring the input when redirecting it? Do you have a code snippet that I could look at?

borna-blazevic:
Hey @andreeaflorescu, sure, here are the 3 variants I tried in Python:
1. Communicate times out
`#!/usr/bin/python

import subprocess
import time

cmd = [
    './vmm-reference/target/debug/vmm-reference',
    '--kernel',
    'path=/home/leonard/Desktop/Other/Rust/vmm_images/vmlinux-focal',
    '--block',
    'path=/home/leonard/Desktop/Other/Rust/vmm_images/rootfs.ext4',
]

proc = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
print('started')
outs, errs = proc.communicate(timeout=15)

print('out = {}'.format(outs))
print('err = {}'.format(errs))`


2. Communicate times out
`#!/usr/bin/python

import subprocess
import time

cmd = [
    './vmm-reference/target/debug/vmm-reference',
    '--kernel',
    'path=/home/leonard/Desktop/Other/Rust/vmm_images/vmlinux-focal',
    '--block',
    'path=/home/leonard/Desktop/Other/Rust/vmm_images/rootfs.ext4',
]

proc = subprocess.Popen(cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
print('started')
time.sleep(10)
outs, errs = proc.communicate(input=bytes('ls', 'utf-8'), timeout=15)

print('out = {}'.format(outs))
print('err = {}'.format(errs))`


3. Works without pipes
`#!/usr/bin/python

import subprocess
import time

cmd = [
    './vmm-reference/target/debug/vmm-reference',
    '--kernel',
    'path=/home/leonard/Desktop/Other/Rust/vmm_images/vmlinux-focal',
    '--block',
    'path=/home/leonard/Desktop/Other/Rust/vmm_images/rootfs.ext4',
]

proc = subprocess.Popen(cmd)
proc.wait()`

Also, I tried it through bash using bash redirection e.g.
`./vmm-reference/target/debug/vmm-reference --kernel path=... --block path=... < pipe_in > pipe_out`


Next week is fine, enjoy your PTO :)

andreeaflorescu:
Hey @borna-blazevic I've tried out the scripts from your comment, but they don't work for me. I get the following error:
```
started
Traceback (most recent call last):
  File "...test_stdin.py", line 46, in <module>
    outs, errs = proc.communicate(input=bytes('ls', 'utf-8'), timeout=15)
TypeError: str() takes at most 1 argument (2 given)
```

We are using pipes when testing the vmm-reference in Python as well, so they should be working. Can you try out the initialization from the Python tests: https://github.com/rust-vmm/vmm-reference/blob/85bcd6e1d73248c775904561c694e593026b53c6/tests/test_run_reference_vmm.py#L89?

This is currently working with Python Pipes, but for the communication with the VMM process we're using the equivalent of `proc.stdin.write` and `proc.stdout.read` to retrieve the output of a certain command. Also, we're setting the STDOUT as non blocking before running any commands. More details in a comment here: https://github.com/rust-vmm/vmm-reference/blob/85bcd6e1d73248c775904561c694e593026b53c6/tests/test_run_reference_vmm.py#L142

andreeaflorescu:
As we discussed offline, it is possible to redirect stdin/stdout to pipes without support added to vmm-reference. This is the preferred approach for now because it is fairly simple to achieve, and it does not need support in the VMM. I am thus closing this PR. If you encounter another blocker, please let us know, and we can reopen it.

Issue #79, opened 2021-02-22
Use log instead of println!/eprintln!


andreeaflorescu:
Import the `log` crate and use it to log needed information.

For now it should be enough to just replace the existing `println!` or `eprintln!`.

tumbleshack:
Hey! 

My friend @seanmichwalsh and I are looking to learn about VVM's and Rust. Mind if we take on this issue? 

It may take us a bit to learn our way around if that's okay with you.

andreeaflorescu:
Sure @tumbleshack , sounds good. If you need any help, let us know.

You can also find us on Slack: https://join.slack.com/t/rust-vmm/shared_invite/enQtODAxMzA2ODIyMTc2LWRhYjIwZmQ0YzUxODJlMTRhZWU2ZDBjYmJiNzBmOWVmYjg4MjY5YWRjYjM0YzQ5YzgyMTBmYzNlMjMzYmZlODU

dodsonmg:
Is this still open?  I'm happy to pick up where #114 left off...

andreeaflorescu:
@dodsonmg yes, this is still opened. We should though reduce the scope for the beginning so that's easier to get it merged. We can do this initially for the `main.rs`, and the `vmm` and `api` crates.

dodsonmg:
Cool.  Thanks @andreeaflorescu.  I'll pick it up.

dodsonmg:
Do you want the log level and log file to be command line arguments, or do you just want default log level settings and log output to go to stderr?

If the former, the log level and file can't be established until we've processed command line arguments, and we'll miss some of the `println!` and `eprintln!` statements that come before then.  This is how cloud hypervisor does it.

If the latter, then I think we can just initialise the logger first and sweep everything up.

andreeaflorescu:
I think how it worked in Firecracker is that we were using `println` and `eprintln` until the logger is initialized, and `error` `info` afterwards. This way you don't loose any logs because you're redirecting the output to a file anyway. I think we could start with something simple where we assume stdout and stderr because I believe these can also just directly be forwarded to files without us doing anything special in the VMM code.

PR #114, opened 2021-03-25
Updated existing prints to make use of log crate

seanmichwalsh:
Imported the `log` and `simple_logger` crates and updated existing print statements to use it in reference to this [issue](https://github.com/rust-vmm/vmm-reference/issues/79). Supported macros include `error!`, `warn!`, `info!`, `debug!`, and `trace!`. The custom `debug!` macro used previously has been replaced with the equivalent in log. Logging level is set in `main.rs` based on the `debug_assertions` read in. When running in release mode, only macros `error!`, `warn!`, and `info!` will print information.

This is the first time tumbleshack and I have contributed to this project. We've tried to be consistent with existing conventions, but if we've made a mistake please let us know - we're happy to learn and fix it.

Signed-off-by: Sean Walsh <seanmichwalsh@protonmail.ch>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,8 @@ license = "Apache-2.0 OR BSD-3-Clause"
 [dependencies]
 vmm = { path = "src/vmm" }
 api = { path = "src/api" }
+log = "0.4.14"
+simple_logger = "1.11.0"
 
 [workspace]
 
--- a/src/api/Cargo.toml
+++ b/src/api/Cargo.toml
@@ -6,5 +6,7 @@ edition = "2018"
 
 [dependencies]
 clap = "2.33.3"
+log = "0.4.14"
+simple_logger = "1.11.0"
 
 vmm = { path = "../vmm" }
--- a/src/api/src/lib.rs
+++ b/src/api/src/lib.rs
@@ -8,6 +8,8 @@
 #![deny(missing_docs)]
 use std::result;
 
+use log::error;
+
 use clap::{App, Arg};
 use vmm::VMMConfig;
 
@@ -62,7 +64,7 @@ impl CLI {
         let help_msg = String::from_utf8_lossy(&help_msg_buf);
 
         let matches = app.get_matches_from_safe(cmdline_args).map_err(|e| {
-            eprintln!("{}", help_msg);
+            error!("{}", help_msg);
             format!("Invalid command line arguments: {}", e)
         })?;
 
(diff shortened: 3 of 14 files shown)


andreeaflorescu:
Hey @seanmichwalsh thanks for your contribution!

Since we are adding new dependencies, we will also need to update the Cargo.lock file.
As a high level comment, we try to keep the dependencies at a bare minimum. With that in mind, I was thinking whether we can in fact just drop the simple_logger dependency. Do you think that would be feasible?

tumbleshack:
Hi @andreeaflorescu , on the simple logger comment: it's definitely feasible to drop it and write that component ourselves. That said, the crate is very small: https://github.com/borntyping/rust-simple_logger

It's source is just one file of around 300 lines. If you'd prefer to leave it out, we can move that code into our own `lib.rs` file

andreeaflorescu:
> Hi @andreeaflorescu , on the simple logger comment: it's definitely feasible to drop it and write that component ourselves. That said, the crate is very small: https://github.com/borntyping/rust-simple_logger
> 

@tumbleshack even though `simple_logger` is very tiny, it is pulling in 8 transitive dependencies. You can see the newly added dependencies by looking at the `Cargo.lock` file. I was thinking it should be possible to get away with less dependencies.

> Its source is just one file of around 300 lines. If you'd prefer to leave it out, we can move that code into our own `lib.rs` file

Pulling in the code from that dependency into our own `lib.rs` is not what I had in mind. We can look at how we can build the most minimal initialization of a log on top of the `log` crate. I haven't looked at the simple logger crate, but that might be a good starting point as well. Otherwise we can also check:
- Cloud Hypervisor: https://github.com/cloud-hypervisor/cloud-hypervisor/blob/9dbea4fe9cd1ee56463e21158358ff1f334c5a41/src/main.rs#L67
- Firecracker: https://github.com/firecracker-microvm/firecracker/blob/master/src/logger/src/logger.rs

tumbleshack:
You got it @andreeaflorescu ; we have spent the last couple of days fighting transitive dependencies on a different Rust project and understand the desire to avoid them. 

We'll check out Cloud Hypervisor and Firecracker logging and let you know here if we have any questions.

lauralt:
Hi, @seanmichwalsh, @tumbleshack! How are things going? Do you need help with anything?

andreeaflorescu:
Closing this due to lack of activity. If you want to start working on this again, feel free to reopen the PR.

Issue #99, opened 2021-03-03
Handle errors gracefully in main()


lauralt:
We are currently panicking if `VMM::try_from()` or `vmm.run()` fail. We could instead handle those errors gracefully (by logging an error, exiting the process etc.).

jcvenegas:
I can help with this

jcvenegas:
I have opened #122  , there I add some handling using two external crates, `thiserror` and `anyhow`.  There I refactor the main binary crate,  given that is using some external creates is OK? or to much  as I dont see to many external creates used, I can just limit to handle the unwrap in main if   thiserror and anyhow is something is not wanted.

lauralt:
Hi, @jcvenegas! You're right, we are indeed trying to limit the number of external dependencies. I think it should be fine to just handle the unwraps, without adding `thiserror` etc. (an example in [this commit](https://github.com/firecracker-microvm/firecracker/commit/cb9a298fd29a9de1385078a0265f932b631b8090)). For now we could just use `println!`s, since there is also [this PR](https://github.com/rust-vmm/vmm-reference/pull/114) who is replacing them with the `log` crate.
Thanks for helping with this!

jcvenegas:
@lauralt I will send a second PR as al alternative proposal

PR #122, opened 2021-05-24
main: Handle errors gracefully in main()

jcvenegas:
Handle calls to `CLI`, `VMM::try_from()`  and` vmm.run()`.

This commit adds `anyhow` and `thiserror` crates to
handle errors.

The `anyhow` crate is used in the main binary to add
extra context and give a pretty view of error trace.

The create `thiserror` was used to easily
derive `std::error::Error` for errors provided by libs.

The new crates pulls the following extra crates:
```diff
+name = "proc-macro2"
+name = "quote"
+name = "syn"
+name = "thiserror-impl"
+name = "unicode-xid"
```

New error format:
examples:
```sh
$ vmm-reference --kernel \
    path=/path/vmlinuz-5.10.25
Error: Failed to create VMM from configurations

Caused by:
    Error issuing an ioctl to KVM.

$ echo $?
1
```

```sh
$ vmm-reference --kernel path
Error: Failed to parse CLI options

Caused by:
    Failed to parse cli Invalid input for kernel: Missing required
    argument: path
```

```
$vmm-reference
Error: Failed to parse CLI options

Caused by:
    Failed to parse cli error: The following required arguments were not provided:
        --kernel <kernel>

    USAGE:
        vmm-reference [OPTIONS] --kernel <kernel>

    For more information try --help
```

Fixes: #99

Signed-off-by: Carlos Venegas <jose.carlos.venegas.munoz@intel.com>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,10 +10,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "anyhow"
+version = "1.0.40"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "28b2cd92db5cbd74e8e5028f7e27dd7aa3090e89e4f2a197cc7c8dfb69c7063b"
+
+[[package]]
 name = "api"
 version = "0.1.0"
 dependencies = [
  "clap",
+ "thiserror",
  "vmm",
 ]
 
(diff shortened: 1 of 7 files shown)


andreeaflorescu:
We typically do not want to add other dependencies, but I find `anyhow` and `thiserror` to reduce quite a few of the burden of dealing with errors, so I am inclined to accept this PR. I also noticed that these are also used by Crosvm & Cloud Hypervisor. As long as we have this dependency only in the VMM glue code, I think it's fine.

@alexandruag @lauralt what do you think?

PR #146, opened 2021-07-03
Explicitly add rw to cmdline

shpark:
Ubuntu (bzimage, ext4) VM boot time takes too long since the file system is mounted read-only.
As a workaround, this patch enforces `rw` to the command line if the virtio-blk is not configured to be read-only

Related to #46


shpark:
Force-pushed to pass commit style test

Issue #124, opened 2021-05-25
Build PE image for CI


andreeaflorescu:
Update the build scripts to build an image on aarch64 as well. We also need to create a PE image build that halts after booting such that we can have the same unit and integration tests implemented on both platforms.

Once this is done, we need to update the [S3 manifest](https://github.com/rust-vmm/vmm-reference/blob/master/tests/tools/s3/resource_manifest.json), and upload the images & the scripts to the S3 bucket used for testing.

For this task we need to update the scripts that build resources: https://github.com/rust-vmm/vmm-reference/tree/master/resources

lauralt:
Fixed by #158

PR #149, opened 2021-07-12
Adjustments needed in make_kernel for using it in Firecracker

dianpopa:
Update the `make_kernel` script to also support the PE format which is what Firecracker uses for arm64 microVMs.

### DIFF
--- a/resources/kernel/make_kernel.sh
+++ b/resources/kernel/make_kernel.sh
@@ -107,14 +107,24 @@ make_initramfs() {
 }
 
 # Usage: validate_kernel_format format
-# Prints the lowercase format name, if one of "elf" or "bzimage".
+# Prints the lowercase format name, if one of "elf" or "bzimage"
+# for x86 or "pe" for aarch64.
 # Exits with error if any other format is specified.
 validate_kernel_format() {
     format="$1"
+    arch=$(uname -m)
 
     kernel_fmt=$(echo "$format" | tr '[:upper:]' '[:lower:]')
-    if [ "$kernel_fmt" != "elf" ] && [ "$kernel_fmt" != "bzimage" ]; then
-        die "Invalid kernel binary format: $kernel_fmt."
+    if [ $arch = "x86_64" ]; then
+        if [ "$kernel_fmt" != "elf" ] && [ "$kernel_fmt" != "bzimage" ]; then
+            die "Invalid kernel binary format: $kernel_fmt for this type of architecture."
+        fi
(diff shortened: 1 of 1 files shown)


resources/disk/make_rootfs.sh
```diff
@@ -37,11 +37,12 @@ Options:
   -d debdir         Directory containing .deb packages for the Linux image.
   -w workdir        Working directory for the kernel build.
   -o diskfile       Name of the resulting disk file.
+  -f flavour        The Ubuntu flavour desired. Defaults to "focal"
```
andreeaflorescu, line 40: Focal describes the version of ubuntu, not the flavor. Can you rename this to version?

andreeaflorescu:
Looks like there was a problem with the CI. I retriggered it now, hopefully it will work :crossed_fingers:

resources/disk/make_rootfs.sh
dianpopa: Actually this commit does not pertain to this PR. I will make sure to update this in this one: #150.

dianpopa:
@andreeaflorescu can you please review again? Also, do you think you can help me find a second reviewer? Thanks!

andreeaflorescu:
@lauralt can you also take a look?

lauralt:
> @lauralt can you also take a look?

Yes, I'll have a look as well.

resources/kernel/make_kernel.sh
```diff
@@ -111,10 +111,19 @@ make_initramfs() {
 # Exits with error if any other format is specified.
 validate_kernel_format() {
```
lauralt, line 112: The description above should be updated.

lauralt (commented):
Can you please make the commit message a bit clearer? I think the arm support involves more changes, e.g. https://github.com/rust-vmm/vmm-reference/issues/124.

resources/kernel/make_kernel.sh
dianpopa: Indeed, updated it.

dianpopa: > Can you please make the commit message a bit clearer? I think the arm support involves more changes, e.g. #124.

Indeed arm support means more in the vmm-reference context. For Firecracker we need just these changes for the moment. I updated the commit message.

dianpopa (commented):
@lauralt PTAL! Thanks!

andreeaflorescu (approved):

lauralt (approved):

PR #158, opened 2021-09-14
Add PE image support in scripts

catalindumitru:
Created a .config file for aarch64 and added the `pe` option throughout the scripts.


tests/tools/s3/resource_manifest.json
```diff
@@ -51,5 +51,132 @@
...
+      "image_format": "elf",
+      "halt_after_boot": false,
+      "distro": "Ubuntu"
+    },
+    {
+      "resource_type": "disk",
+      "resource_name": "ubuntu-focal-rootfs.ext4",
+      "relative_path": "disk/",
+      "distro": "Ubuntu",
+      "compatible_kernels": "v1/kernel/bzimage/bzimage-focal|v1/kernel/elf/vmlinux-focal"
```
andreeaflorescu, line 179: Can you also add the following intird configuration:
```
    {
      "resource_type": "initrd",
      "resource_name": "aarch64-initramfs.cpio",
      "relative_path": "kernel/pe"
    }
```

catalindumitru: Added it.

lauralt: Should we add it to v2 as well?

lauralt:
We should also mention about the kernel version update in the first commit message.

tests/tools/s3/resource_manifest.json
andreeaflorescu: I think we shouldn't. I'll remove it from the v2 directory in S3, that was a mistake on my part, sorry.

lauralt (dismissed):

andreeaflorescu (approved):

lauralt (dismissed):

tests/tools/s3/resource_manifest.json
```diff
@@ -51,5 +51,137 @@
...
```
lauralt, line 112: It is compatible with v2 ones as well, so we could probably update this; same comment for v3.

andreeaflorescu (approved):

lauralt (approved):

PR #162, opened 2021-09-24
rename high_mem_start to load_addr

andreeaflorescu:
This patch is in preparation of adding aarch64 support. Naming the
field load_addr makes it a bit more platform independent.

### DIFF
--- a/docs/DESIGN.md
+++ b/docs/DESIGN.md
@@ -151,7 +151,7 @@ effort.
 vmm-reference                                                           \
     --memory size_mib=1024                                          \
     --vcpu num=1                                                 \
-    --kernel path=/path/to/vmlinux,himem_start=1024,cmdline="pci=off"   \
+    --kernel path=/path/to/vmlinux,kernel_load_addr=1024,cmdline="pci=off"   \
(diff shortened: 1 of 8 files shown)


alexandruag (approved):

docs/DESIGN.md
```diff
@@ -151,7 +151,7 @@ effort.
 vmm-reference                                                           \
     --memory size_mib=1024                                          \
     --vcpu num=1                                                 \
-    --kernel path=/path/to/vmlinux,himem_start=1024,cmdline="pci=off"   \
+    --kernel path=/path/to/vmlinux,kernel_load_addr=1024,cmdline="pci=off"   \
```
lauralt, line 154: If `kernel_load_addr` belongs to `kernel` argument, would it be enough to name it simply `load_addr`?

lauralt (approved):

docs/DESIGN.md
andreeaflorescu: In the same configuration we might also add the value for `cmdline_load_addr`. That's why I added kernel as well. It is a bit weird indeed.

PR #165, opened 2021-09-30
update make_test_resources.sh

lauralt:
The reason for this PR is to make it easier to upload new versions of resources to S3
by storing the resources in the same place where the tests are downloading them. We can
also have now full reproducible images based on the container version.


resources/CONTAINER_VERSION
```diff
@@ -0,0 +1 @@
+LATEST=v13
```
andreeaflorescu, line 1: It would be easier to create a script that parses it if we just write `13` instead of `LATEST=13`.

resources/make_test_resources.sh
```diff
@@ -12,24 +12,32 @@ SOURCE=$(readlink -f "$0")
...
-    ./kernel/make_kernel_image_deb.sh -f elf -k vmlinux-focal -w $DEB_DIR -j 2
-    ./kernel/make_kernel_image_deb.sh -f elf -k vmlinux-focal -w $DEB_DIR -j 2 -h
-    ./kernel/make_kernel_image_deb.sh -f bzimage -k bzimage-focal -w $DEB_DIR -j 2
-    ./kernel/make_kernel_image_deb.sh -f bzimage -k bzimage-focal -w $DEB_DIR -j 2 -h
-    ./disk/make_rootfs.sh -d /tmp/ubuntu-focal/linux-5.4.81/deb/ -w $DISK_DIR -o rootfs.ext4
+    ./kernel/make_kernel_busybox_image.sh -f elf -k vmlinux-hello-busybox -w $TMP_KERNEL_DIR -j 2
+    ./kernel/make_kernel_busybox_image.sh -f elf -k vmlinux-hello-busybox -w $TMP_KERNEL_DIR -j 2 -h
+    ./kernel/make_kernel_busybox_image.sh -f bzimage -k bzimage-hello-busybox -w $TMP_KERNEL_DIR -j 2
+    ./kernel/make_kernel_busybox_image.sh -f bzimage -k bzimage-hello-busybox -w $TMP_KERNEL_DIR -j 2 -h
+    mv $TMP_KERNEL_DIR/linux-5.4.81/vmlinux-hello-busybox $TMP_KERNEL_DIR/linux-5.4.81/vmlinux-hello-busybox-halt \
```
andreeaflorescu, line 29: Can you separate these 2 by a whitespace so it's more clear that you are building them, and then move them to where they are expected to be found by the vmm-reference tests? It would also be nice to add a comment that we're moving the resources to make it easier to upload the whole `resource` folder to S3.

resources/CONTAINER_VERSION
lauralt: Updated. I added `LATEST` to make it clearer, but if it's not necessary, we can have only the number here.

resources/make_test_resources.sh
lauralt: Yep, sorry, added more comments.

resources/make_test_resources.sh
```diff
@@ -12,24 +12,42 @@ SOURCE=$(readlink -f "$0")
 TEST_RESOURCE_DIR="$(dirname "$SOURCE")"
 cd $TEST_RESOURCE_DIR
 
-# If user doesn't provide a value for a variable, use the default one (which is
-# the one used by the tests).
-KERNEL_DIR=${KERNEL_DIR:="/tmp/vmlinux_busybox"}
-DEB_DIR=${DEB_DIR:="/tmp/ubuntu-focal"}
-DISK_DIR=${DISK_DIR:="/tmp/ubuntu-focal-disk"}
+# If user doesn't provide a value for a variable, use the default one.
```
catalindumitru, line 15: Nit: "the user" 😆

catalindumitru (dismissed):

andreeaflorescu (approved):

alexandruag (approved):

PR #166, opened 2021-10-01
Use `linux_loader::Cmdline` instead of `String`

### LINKED ISSUE #98: Use Cmdline from linux_loader instead of String
Figure out the most straightforward way to reuse `Cmdline` from the `linux-loader` crate instead of `String` for building/passing/handling parts of the kernel command line that we're passing to the guest (also double check there are no concerns or missing features beforehand).

alexandruag:
This PR replaces the use of `String` with `linux_loader::Cmdline` for the `cmdline` member of the `KernelConfig` structure.

Fixes #98.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -121,9 +121,9 @@ checksum = "8916b1f6ca17130ec6568feccee27c156ad12037880833a3b842a823236502e7"
 
 [[package]]
 name = "linux-loader"
-version = "0.3.0"
+version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c819cc8275b0f2c1ed9feec455ca288b45d82932384a6a5f7a86812ee3427459"
+checksum = "8a5e77493808403a6bd56a301a64ea6b9342e36ea845044bf0dfdf56fe52fa08"
 dependencies = [
  "vm-memory",
 ]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,9 +207,9 @@ source = "git+https://github.com/rust-vmm/vm-device?rev=5847f12#5847f1286492b719
 
 [[package]]
 name = "vm-memory"
-version = "0.5.0"
+version = "0.6.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


Cargo.lock
```diff
@@ -1,5 +1,7 @@
 # This file is automatically @generated by Cargo.
 # It is not intended for manual editing.
+version = 3
+
```
andreeaflorescu, line 4: Are you using a newer Rust/cargo version? I didn't see this change locally.

Cargo.toml
```diff
@@ -19,3 +19,6 @@ codegen-units = 1
 lto = true
 panic = "abort"
 
+[patch.crates-io]
+# Using this patch until a version > 4.0 gets published.
```
andreeaflorescu, line 23: Can we mark it with a `TODO` instead so that we can keep track of it?

src/vmm/src/config/mod.rs
```diff
@@ -158,11 +175,17 @@ impl TryFrom<&str> for KernelConfig {
...
+
+        let cmdline_str = arg_parser
             .value_of("cmdline")
             .map_err(ConversionError::new_kernel)?
             .unwrap_or_else(|| DEFAULT_KERNEL_CMDLINE.to_string());
 
+        let mut cmdline = Cmdline::new(KERNEL_CMDLINE_CAPACITY);
+        if cmdline.insert_str(cmdline_str).is_err() {
+            return Err(ConversionError::new_kernel("Kernel cmdline capacity error"));
+        }
```
andreeaflorescu, line 187: Can we use `map_err` for consistency reasons?

Cargo.lock
alexandruag: Yes, it was from a newer version; reverted it.

Cargo.toml
alexandruag: done

src/vmm/src/config/mod.rs
alexandruag: sure

andreeaflorescu (approved):

lauralt (approved):
nit: would be nice to mention in the last commit message that it fixes #98.

andreeaflorescu (approved):

lauralt (approved):

PR #167, opened 2021-10-08
First steps towards refactoring the Virtio device model

alexandruag:
This PR brings the first of a set of changes that aims to simplify the Virtio device model of the reference VMM, reduce the amount of boilerplate code required for each individual device, and make it easier to have devices that are clearly separated from the specific VMM that uses them and become generic as long as certain assumptions are met.

The fist commit introduces two traits that model the requirements devices expect from the environment in order to perform their setup and operate, in a way which does not make any additional assumptions about the VMM. Things are mostly Virtio over MMIO specific for now, but can be adapted to support the PCI transport as well in the future.

The second commit introduces the `RefVirtioDevice` abstraction, which contains logic that is common for all devices. Specific functionality is added by plugging in an inner object, which implements the `RefVirtioDeviceT` trait according to the characteristics of every particular device (i.e. block, net, etc.) Finally, the last commit showcases how the above abstractions can be used as part of the block device implementation. Future PRs will bring additional refactoring.

### DIFF
--- a/src/devices/src/virtio/block/device.rs
+++ b/src/devices/src/virtio/block/device.rs
@@ -1,114 +1,71 @@
 // Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
-use std::borrow::{Borrow, BorrowMut};
 use std::fs::OpenOptions;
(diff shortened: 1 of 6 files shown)


andreeaflorescu:
Closing this, we'll be implementing this as part of vm-virtio

Issue #159, opened 2021-09-20
remove support for busybox 1.26


andreeaflorescu:
Let's switch to the newest available busybox version as this comes with additional tools that we can use during testing.

For now, this is blocked on publishing a new container version https://github.com/rust-vmm/rust-vmm-container/issues/53

PR #168, opened 2021-10-11
Remove support for busybox 1.26

gsserge:
Fixes: https://github.com/rust-vmm/vmm-reference/issues/159

Signed-off-by: Sergii Glushchenko <gsserge@amazon.com>

### DIFF
--- a/resources/kernel/configs.sh
+++ b/resources/kernel/configs.sh
@@ -3,9 +3,9 @@
 # Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 
-# This scripts determines the busybox version to be used based on the
-# current glibc version (1.32.1 for glibc version > 2.31)
-# Also kernel version and config parameters are exported from here.
+# The busybox version, kernel version and config parameters are exported
+# from this script. The current busybox version is compatible with glibc
+# version > 2.31.
 
 arch=$(uname -i)
 KERNEL_VERSION="5.4.81"


.buildkite/build_resources.py
```diff
@@ -0,0 +1,30 @@
...
+    try:
+        conn = http.client.HTTPSConnection("api.github.com", timeout=2)
+        path = f"/repos/rust-vmm/vmm-reference/pulls/{pull_request}/files"
+        conn.request('GET', path, headers={'user-agent': 'py.http/3'})
+        content = conn.getresponse().read().decode('utf-8')
+        for v in json.loads(content):
+            if v['filename'].startswith('resources/'):
+                pipeline['steps'][0]['skip'] = False
+    except Exception as e:
+        pass
```
andreeaflorescu, line 28: I think it might be a good idea to print the exception. It might help with debugging in case things are not working as expected.

andreeaflorescu (approved):

andreeaflorescu (dismissed):

.buildkite/build_resources.py
```diff
@@ -0,0 +1,30 @@
...
+import json
+
+pipeline = {'steps': [
+    {'label': 'trigger-build-resources',
+     'trigger': 'vmm-reference-build-resources',
+     'skip': 'No changes in resources or not a Pull Request'}
+]}
+
+pull_request = os.environ.get('BUILDKITE_PULL_REQUEST')
+if pull_request and pull_request.isnumeric():
```
lauralt, line 18: Can we add a comment here saying that we do the `isnumeric()` check because `pull_request` will be "false" if it's not a pull request or something like that?

PR #184, opened 2021-11-02
Prepare gdt for upstream

andreeaflorescu:
This PR improves the GDT code to make it ready to be upstreamed. The approach is to move the code from vm-vcpu (which will contain the product specific code) to vm-vcpu-ref (which is the crate that's going to be published).
Then, future improvements are added directly in the vm-vcpu-ref code.

For some weird reason in tests I had to explicitly write the type of guest_memory to be `GuestMemoryMmap` because otherwise the tests will not compile with the error being that the `B` type corresponding to the Bitmap is missing.

### DIFF
--- a/src/vm-vcpu-ref/src/x86_64/gdt.rs
+++ b/src/vm-vcpu-ref/src/x86_64/gdt.rs
@@ -5,126 +5,287 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the THIRD-PARTY file.
 
-// For GDT details see arch/x86/include/asm/segment.h
+//! The `gdt` module provides abstractions for building a Global Descriptor Table (GDT).
+//!
+//! These abstractions are built to resemble the structures defined in the
+//! [Intel Manual](https://www.intel.com/content/dam/www/public/us/en/documents/manuals/64-ia-32-architectures-software-developer-vol-3a-part-1-manual.pdf).
 
 use kvm_bindings::kvm_segment;
-use vm_memory::{Bytes, GuestAddress, GuestMemory};
+use vm_memory::{ByteValued, Bytes, GuestAddress, GuestMemory, GuestMemoryError};
 
 use std::mem;
 
+/// The offset at which GDT resides in memory.
 pub const BOOT_GDT_OFFSET: u64 = 0x500;
+/// The offset at which IDT resides in memory.
 pub const BOOT_IDT_OFFSET: u64 = 0x520;
-pub const BOOT_GDT_MAX: usize = 4;
+/// Maximum number of GDT entries as defined in the Intel Specification.
+pub const MAX_GDT_SIZE: usize = 1 << 13;
 
-pub fn gdt_entry(flags: u16, base: u32, limit: u32) -> u64 {
-    ((u64::from(base) & 0xff00_0000u64) << (56 - 24))
-        | ((u64::from(flags) & 0x0000_f0ffu64) << 40)
-        | ((u64::from(limit) & 0x000f_0000u64) << (48 - 16))
-        | ((u64::from(base) & 0x00ff_ffffu64) << 16)
-        | (u64::from(limit) & 0x0000_ffffu64)
+/// Errors associated with operations on the GDT.
+#[derive(Debug)]
+pub enum Error {
+    /// Invalid memory access.
+    GuestMemory(GuestMemoryError),
+    /// Too many entries in the GDT.
+    TooManyEntries,
 }
 
-pub fn get_base(entry: u64) -> u64 {
-    (((entry) & 0xFF00_0000_0000_0000) >> 32)
-        | (((entry) & 0x0000_00FF_0000_0000) >> 16)
-        | (((entry) & 0x0000_0000_FFFF_0000) >> 16)
-}
-
-pub fn get_limit(entry: u64) -> u32 {
-    ((((entry) & 0x000F_0000_0000_0000) >> 32) | ((entry) & 0x0000_0000_0000_FFFF)) as u32
-}
-
-pub fn get_g(entry: u64) -> u8 {
-    ((entry & 0x0080_0000_0000_0000) >> 55) as u8
-}
-
-pub fn get_db(entry: u64) -> u8 {
-    ((entry & 0x0040_0000_0000_0000) >> 54) as u8
-}
-
-pub fn get_l(entry: u64) -> u8 {
-    ((entry & 0x0020_0000_0000_0000) >> 53) as u8
-}
-
-pub fn get_avl(entry: u64) -> u8 {
-    ((entry & 0x0010_0000_0000_0000) >> 52) as u8
-}
-
-pub fn get_p(entry: u64) -> u8 {
-    ((entry & 0x0000_8000_0000_0000) >> 47) as u8
-}
-
-pub fn get_dpl(entry: u64) -> u8 {
-    ((entry & 0x0000_6000_0000_0000) >> 45) as u8
-}
-
-pub fn get_s(entry: u64) -> u8 {
-    ((entry & 0x0000_1000_0000_0000) >> 44) as u8
-}
-
-pub fn get_type(entry: u64) -> u8 {
-    ((entry & 0x0000_0F00_0000_0000) >> 40) as u8
-}
-
-/// Automatically build the kvm struct for SET_SREGS from the kernel bit fields.
-///
-/// # Arguments
-///
-/// * `entry` - The gdt entry.
-/// * `table_index` - Index of the entry in the gdt table.
-pub fn kvm_segment_from_gdt(entry: u64, table_index: u8) -> kvm_segment {
-    kvm_segment {
-        base: get_base(entry),
-        limit: get_limit(entry),
-        selector: u16::from(table_index * 8),
-        type_: get_type(entry),
-        present: get_p(entry),
-        dpl: get_dpl(entry),
-        db: get_db(entry),
-        s: get_s(entry),
-        l: get_l(entry),
-        g: get_g(entry),
-        avl: get_avl(entry),
-        padding: 0,
-        unusable: match get_p(entry) {
-            0 => 1,
-            _ => 0,
-        },
+impl From<GuestMemoryError> for Error {
+    fn from(inner: GuestMemoryError) -> Self {
+        Error::GuestMemory(inner)
     }
 }
 
-pub fn write_gdt_table<T: GuestMemory>(
-    table: &[u64],
-    guest_mem: &T,
-) -> std::result::Result<(), vm_memory::GuestMemoryError> {
-    let boot_gdt_addr = GuestAddress(BOOT_GDT_OFFSET);
-    for (index, entry) in table.iter().enumerate() {
-        let addr = guest_mem
-            .checked_offset(boot_gdt_addr, index * mem::size_of::<u64>())
-            .ok_or(vm_memory::GuestMemoryError::InvalidGuestAddress(
-                boot_gdt_addr,
-            ))?;
-        guest_mem.write_obj(*entry, addr)?;
+/// Results corresponding to operations on the GDT.
+pub type Result<T> = std::result::Result<T, Error>;
+
+#[derive(Copy, Clone, Default, Debug)]
+/// A segment descriptor is a data structure in a GDT (Global Descriptor Table)
+/// that provides the processor with the size and location of a segment, as
+/// well as access control and status information.
+///
+/// The segment descriptor is referred to in the Linux Kernel as a GDT entry.
+/// More information about the segment descriptor, its inner structure, and
+/// types of segment descriptors is available in Section "3.4.5 Segment
+/// Descriptors" of the Intel Manual Volume 3a.
+pub struct SegmentDescriptor(pub u64);
+
+// Safe because SegmentDescriptor is just a wrapper over u64.
+unsafe impl ByteValued for SegmentDescriptor {}
+
+impl SegmentDescriptor {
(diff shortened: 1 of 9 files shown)


andreeaflorescu:
Note for reviewers: the PR is easier to review commit by commit because in the first commit I'm just moving the `gdt` from vm-vcpu to vm-vcpu-ref, and in the following commit I'm actually doing the work needed for publishing and some updates to the interface.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+    ///
+    /// let guest_memory: GuestMemoryMmap =
+    ///     GuestMemoryMmap::from_ranges(&[(GuestAddress(0), 1024 << 20)]).unwrap();
+    /// let gdt_table = Gdt::default();
+    /// gdt_table.write_to_mem(&guest_memory).unwrap();
+    /// ```
+    pub fn write_to_mem<Memory: GuestMemory>(
+        &self,
+        mem: &Memory,
+    ) -> std::result::Result<(), vm_memory::GuestMemoryError> {
```
gsserge, line 210: Full `std::result::` path is probably not needed here.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+        let boot_gdt_addr = GuestAddress(BOOT_GDT_OFFSET);
+        for (index, entry) in self.iter().enumerate() {
+            let addr = mem
+                .checked_offset(boot_gdt_addr, index * mem::size_of::<u64>())
```
gsserge, line 214: Is it possible to use `size_of::<SegmentDescriptor>` instead of `u64` to make it more clear?

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+                .ok_or(vm_memory::GuestMemoryError::InvalidGuestAddress(
```
gsserge, line 215: Somehow `clippy` always complains that I should use `.ok_or_else(|| {...})` with a closure instead of creating an error value regardless of the control flow :)

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+    /// ```rust
+    /// use vm_vcpu_ref::x86_64::gdt::SegmentDescriptor;
+    ///
+    /// let code_descriptor = SegmentDescriptor::from(0xa09b, 0, 0xfffff);
+    /// // The code descriptor is found at position 1 in the GDT.
+    /// // More examples are available here: https://wiki.osdev.org/GDT_Tutorial
+    /// let table_index = 1;
+    /// let code_segment = code_descriptor.create_kvm_segment(table_index);
+    /// ```
+    pub fn create_kvm_segment(&self, table_index: u8) -> kvm_segment {
```
gsserge, line 140: Where is the aforementioned `entry` argument?

andreeaflorescu: The entry is now `self`. I'll update the documentation.

src/vm-vcpu-ref/src/x86_64/gdt.rs
andreeaflorescu: I'll include `std::result`

src/vm-vcpu-ref/src/x86_64/gdt.rs
andreeaflorescu: Will do.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -183,19 +180,16 @@ impl KvmVcpu {
...
-            gdt_entry(0xa09b, 0, 0xfffff), // CODE
-            gdt_entry(0xc093, 0, 0xfffff), // DATA
-            gdt_entry(0x808b, 0, 0xfffff), // TSS
-        ];
+        let gdt_table = Gdt::default();
 
-        let code_seg = kvm_segment_from_gdt(gdt_table[1], 1);
-        let data_seg = kvm_segment_from_gdt(gdt_table[2], 2);
-        let tss_seg = kvm_segment_from_gdt(gdt_table[3], 3);
+        let code_seg = gdt_table[1].create_kvm_segment(1);
```
lauralt, line 185: I am wondering if it would make sense to have this method as part of `Gdt`, and avoid this way repeating the index. Looks like the `table_index` would always be the same as the index from the `Gdt` (or not?).

Basically something like:

```rust
impl Gdt {
    pub fn create_kvm_segment(&self, table_index: u8) -> kvm_segment {
       let entry = self.get(table_index as usize).ok_or(...);
       ...
    }
}
```

I think this way we could also validate in `create_kvm_segment` that the table_index is valid and return an error if not, as opposed to `gdt_table[x]` which would panic.

src/vm-vcpu-ref/src/x86_64/gdt.rs
andreeaflorescu: `ok_or` is eagerly evaluated, while `ok_or_else` is lazy evaluated. In this particular case we're not calling any functions, so I think we should be okay with using `ok_or`.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -36,7 +35,7 @@ impl SegmentDescriptor {
     /// based on base, flags, and limit.
     ///
     /// The segment descriptor can then be used to create the Global Descriptor
-    /// Table (GDT). For more details, check out the [`write_gdt_table`] function.
+    /// Table (GDT). For more details, check out the [Gdt contructor](Gdt::new).
```
lauralt, line 38: contructor -> constructor.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+    /// ```rust
+    /// use vm_memory::{GuestAddress, GuestMemoryMmap};
+    /// use vm_vcpu_ref::x86_64::gdt::Gdt;
+    ///
+    /// let guest_memory: GuestMemoryMmap =
+    ///     GuestMemoryMmap::from_ranges(&[(GuestAddress(0), 1024 << 20)]).unwrap();
+    /// let gdt_table = Gdt::default();
+    /// gdt_table.write_to_mem(&guest_memory).unwrap();
+    /// ```
+    pub fn write_to_mem<Memory: GuestMemory>(
```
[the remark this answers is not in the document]
andreeaflorescu: `M` is not very descriptive, I prefer to use types that are more descriptive to what they actually represent.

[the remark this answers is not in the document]
lauralt: Then we should use the same here as well: https://github.com/rust-vmm/vmm-reference/pull/184/files#diff-cf2497bf86bf90ca46af24b2b0d285cea74f3a627e979e8f2a97cedb885f8de6R251 I guess.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: I'll think about this a bit longer. The current abstraction is not ideal, but I am not sure that it helps to add another level of indirection by having this function part of the `Gdt` and having to validate indexes. I'd rather have an assert!(gdt_table.len((), 4) before doing these accesses to avoid any risk of running into a programming error.

I'll come back to this in a bit.

src/vm-vcpu-ref/src/x86_64/gdt.rs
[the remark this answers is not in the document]
andreeaflorescu: I didn't update that function at all, so that's why I didn't spot it. I can update it as well.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+    /// gdt.push(SegmentDescriptor::from(0x92, 0, 0xffffffff));
+    /// gdt.push(SegmentDescriptor::from(0x89, 0, 0xffffffff));
+    /// ```
+    pub fn new() -> Gdt {
+        Gdt(vec![])
+    }
+
+    /// Writes the GDT (Global Descriptor Table) into Guest Memory.
+    ///
+    /// # Example - Creating the GDT for Linux
```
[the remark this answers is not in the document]
andreeaflorescu: For this one I find it superfluous to have the arguments section because it would just repeat what the documentation is already saying. I excluded it on purpose, but I can add it if you feel strongly about it.

src/vm-vcpu/src/vcpu/mod.rs
lauralt: I don't particularly like the fact that you have to be careful to pass the same index twice, and it seemed like a good opportunity to avoid this with the `Gdt` abstraction. We could also have an assert in `create_kvm_segment` for the index validation and then use self[table_index as usize]. But I'm fine anyway, it was just a suggestion.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
-}
+#[derive(Copy, Clone, Default, Debug)]
+/// A segment descriptor is a data structure in a GDT (Global Descriptor Table)
+/// that provides the processor with the size and location of a segment, as
+/// well as access control and status information.
+///
+/// The segment descriptor is referred to in the Linux Kernel as a GDT entry.
+/// More information about the segment descriptor, its inner structure, and
+/// types of segment descriptors is available in Section "3.4.5 Segment
+/// Descriptors" of the Intel Manual Volume 3a.
```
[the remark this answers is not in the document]
andreeaflorescu: Added link in top level docs.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+        // flags, then there is the Limit (which we need to ignore when setting the
+        // flags, then there are more flags (from bits 52 to 55).
+        // It might make sense in the future to create a wrapper over the flags
+        // so that it's easier to set it; for example, we can have something like:
+        // Flags::set_available(val).set_present(..) and so on. The flags are
+        // defined in Figure 3-8. Segment Descriptor table of the intel manual.
+        // The getters below (like avl, g, l, p) also are for the flags and their offset
+        // is defined in the spec.
+        SegmentDescriptor(
+            ((u64::from(base) & 0xff00_0000u64) << (56 - 24))
```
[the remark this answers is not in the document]
andreeaflorescu: Done.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+    /// * `limit`: A 20-bit value (the most significant 12 bits are ignored from
+    ///            the value) that tells the maximum addressable unit.
+    /// * `flags`: Depending on the segment type, the flags set up various
+    ///            properties such as the privilege level, type, and granularity.
+    ///            The full list of available flags is available in section
+    ///            3.4.5 of the Intel® 64 and IA-32 Architectures Developer's
+    ///            Manual: Vol. 3A.
+    pub fn from(flags: u16, base: u32, limit: u32) -> Self {
+        // The segment descriptor has the following inner structure:
+        // |          Base 0:15                     |          Limit 0:15              |
```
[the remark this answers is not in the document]
andreeaflorescu: Done.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -1,114 +1,254 @@
...
+    }
 
-pub fn get_dpl(entry: u64) -> u8 {
-    ((entry & 0x0000_6000_0000_0000) >> 45) as u8
+    /// Build a `kvm_segment` from a GDT segment descriptor and the
+    /// selector (`table_index`).
+    ///
+    /// # Arguments
+    ///
+    /// * `entry` - The gdt entry.
```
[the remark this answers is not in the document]
andreeaflorescu: Done.

src/vm-vcpu-ref/src/x86_64/gdt.rs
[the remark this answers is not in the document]
andreeaflorescu: Done.

src/vm-vcpu-ref/src/x86_64/gdt.rs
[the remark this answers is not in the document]
andreeaflorescu: I made it such that it cannot panic anymore by adding a MAX_GDT_SIZE and removing the deref.

src/vm-vcpu-ref/src/x86_64/gdt.rs
```diff
@@ -141,5 +302,58 @@ mod tests {
...
+        }
+
+        assert_eq!(gdt.0.len(), MAX_GDT_SIZE);
+        // Pushing one more element returns an error.
+        assert!(gdt.try_push(SegmentDescriptor(0)).is_err());
+    }
+
+    #[test]
+    fn test_memory_constraints() {
+        // The Segment descriptor needs to fit in an u64 otherwise the
```
lauralt, line 356: missing end of phrase.

lauralt (dismissed):

lauralt (approved):

gsserge (approved):

PR #185, opened 2021-11-03
Upstream cpuid

andreeaflorescu:
This PR prepare the `cpuid` module for upstream. The first 3 commits should be ignored as they're part of #184. It was easier to continue the setup on top of those commits.

For the upstream, I've added tests, documentation, and some other tiny updates. More details in each corresponding commit.

### DIFF
--- a/src/vm-vcpu-ref/src/x86_64/cpuid.rs
+++ b/src/vm-vcpu-ref/src/x86_64/cpuid.rs
@@ -15,6 +15,7 @@ const ECX_TSC_DEADLINE_TIMER_SHIFT: u32 = 24; // TSC deadline mode of APIC timer
 const ECX_HYPERVISOR_SHIFT: u32 = 31; // Flag to be set when the cpu is running on a hypervisor.
 const EDX_HTT_SHIFT: u32 = 28; // Hyper Threading Enabled.
 
+/// Filter CPUID.
 pub fn filter_cpuid(kvm: &Kvm, vcpu_id: usize, cpu_count: usize, cpuid: &mut CpuId) {
     for entry in cpuid.as_mut_slice().iter_mut() {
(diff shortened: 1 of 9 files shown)


lauralt (approved):

src/vm-vcpu-ref/src/x86_64/cpuid.rs
```diff
@@ -15,11 +15,32 @@ const ECX_TSC_DEADLINE_TIMER_SHIFT: u32 = 24; // TSC deadline mode of APIC timer
...
+/// fn default_cpuid(cpu_index: usize, num_vcpus: usize) -> Result<CpuId, Error> {
+///     let kvm = Kvm::new()?;
+///     let mut cpuid = kvm.get_supported_cpuid(kvm_bindings::KVM_MAX_CPUID_ENTRIES)?;
+///     filter_cpuid(&kvm, cpu_index, num_vcpus, &mut cpuid);
+///     Ok(cpuid)
+/// }
+///
+/// # default_cpuid(0, 1).unwrap();
+/// ```
 pub fn filter_cpuid(kvm: &Kvm, vcpu_id: usize, cpu_count: usize, cpuid: &mut CpuId) {
```
lauralt, line 40: Would it be better to have vcpu_id and cpu_count as u8 for example, so we would avoid possible unsafe `as u32`s below or isn't truncating the values a problem here?

gsserge (dismissed):

src/vm-vcpu-ref/src/x86_64/cpuid.rs
andreeaflorescu: Done.

gsserge (approved):

lauralt (approved):

PR #186, opened 2021-11-05
Upstream MP Table

andreeaflorescu:
This PR moves the MP table modules to vm-vcpu-ref and then applies changes that are needed for publishing to crates.io. A few things are changed, more details in the commit mesasge.


src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -15,20 +15,19 @@ use kvm_ioctls::{VcpuExit, VcpuFd, VmFd};
...
 use vm_memory::{Address, Bytes, GuestAddress, GuestMemory, GuestMemoryError};
-#[cfg(target_arch = "x86_64")]
-use vm_vcpu_ref::x86_64::gdt::{self, write_idt_value, Gdt, BOOT_GDT_OFFSET, BOOT_IDT_OFFSET};
 use vmm_sys_util::errno::Error as Errno;
 use vmm_sys_util::signal::{register_signal_handler, SIGRTMIN};
 use vmm_sys_util::terminal::Terminal;
 
+use interrupts::*;
 use utils::debug;
+#[cfg(target_arch = "x86_64")]
```
gsserge, line 24: Is it the goal that `#[cfg]` is applied only to `use vm_vcpu_ref::x86_64::gdt::{self, write_idt_value, Gdt, BOOT_GDT_OFFSET, BOOT_IDT_OFFSET};` and not to `use vm_vcpu_ref::x86_64::mptable;`. Or both `use`'s should be conditionally compiled?

andreeaflorescu: They should both be conditionally compiled. It is going to take some time until we're actually able to receive errors for these kind of things :(. I'll update it.

src/vm-vcpu-ref/src/x86_64/mptable.rs
```diff
@@ -112,229 +117,266 @@ fn mpf_intel_compute_checksum(v: &mpspec::mpf_intel) -> u8 {
...
+        mem.read_from(base_mp, &mut io::repeat(0), mp_size)
+            .map_err(|_| Error::Clear)?;
+
+        {
+            let size = mem::size_of::<MpfIntel>() as u64;
+            let mut mpf_intel = MpfIntel(mpspec::mpf_intel {
+                signature: SMP_MAGIC_IDENT,
+                length: 1,
+                specification: MPC_SPEC as u8,
+                physptr: (base_mp.raw_value() + size) as u32,
```
lauralt, line 200: Can we add a comment here explaining why the conversion from u64 to u32 is safe?

src/vm-vcpu-ref/src/x86_64/mptable.rs
```diff
@@ -112,229 +117,266 @@ fn mpf_intel_compute_checksum(v: &mpspec::mpf_intel) -> u8 {
...
-    if let Some(end_mp) = base_mp.checked_add((mp_size - 1) as u64) {
-        if !mem.address_in_range(end_mp) {
-            return Err(Error::NotEnoughMemory);
+    /// Write the MP Table in guest memory.
+    pub fn write<M: GuestMemory>(&self, mem: &M) -> Result<()> {
+        // Used to keep track of the next base pointer into the MP table.
+        let mut base_mp = GuestAddress(MPTABLE_START);
+        let mp_size = self.size();
+        // The checked_add here ensures the all of the following base_mp.unchecked_add's will be without
+        // overflow.
```
lauralt, line 179: I would also add a short comment here saying something like: "all the `as u8/u16` below are safe because the numbers from which the conversions are made have small values".

src/vm-vcpu-ref/src/x86_64/mptable.rs
```diff
@@ -112,229 +117,266 @@ fn mpf_intel_compute_checksum(v: &mpspec::mpf_intel) -> u8 {
...
-            .map_err(|_| Error::WriteMpcTable)?;
-    }
+        // At this point we know the size of the mp_table.
+        let table_end = base_mp;
+
+        {
+            let mut mpc_table = MpcTable(mpspec::mpc_table {
+                signature: MPC_SIGNATURE,
+                // it's safe to use unchecked_offset_from because
+                // table_end > table_base
```
lauralt, line 322: Maybe we can also explain why the `as u16` is safe.

src/vm-vcpu-ref/src/x86_64/mptable.rs
```diff
@@ -112,229 +117,266 @@ fn mpf_intel_compute_checksum(v: &mpspec::mpf_intel) -> u8 {
...
-        assert!(setup_mptable(&mem, num_cpus).is_err());
+        let mptable = MpTable::new(num_cpus).unwrap();
+        let mem: GuestMemoryMmap =
+            GuestMemoryMmap::from_ranges(&[(GuestAddress(MPTABLE_START), mptable.size())]).unwrap();
+        mptable.write(&mem).unwrap();
+
+        // Test mptable does not fit in guest memory.
+        let num_cpus = 5;
+        let mptable = MpTable::new(num_cpus).unwrap();
+        assert_eq!(mptable.write(&mem).unwrap_err(), Error::NotEnoughMemory);
```
lauralt, line 367: Would it be possible to test other error conditions as well, e.g. WriteMpfIntel, WriteMpcCpu?

src/vm-vcpu-ref/src/x86_64/mptable.rs
```diff
@@ -378,24 +418,21 @@ mod tests {
...
     fn cpu_entry_count() {
-        let mem = GuestMemoryMmap::from_ranges(&[(
-            GuestAddress(MPTABLE_START),
-            compute_mp_size(MAX_SUPPORTED_CPUS as u8),
-        )])
-        .unwrap();
+        let mem: GuestMemoryMmap =
+            GuestMemoryMmap::from_ranges(&[(GuestAddress(MPTABLE_START), 1024 << 20)]).unwrap();
 
         for i in 0..MAX_SUPPORTED_CPUS as u8 {
```
lauralt, line 424: nit: MAX_SUPPORTED_CPUS is already an u8.

src/vm-vcpu-ref/src/x86_64/mptable.rs
```diff
@@ -112,229 +117,266 @@ fn mpf_intel_compute_checksum(v: &mpspec::mpf_intel) -> u8 {
...
+            mpspec::MP_BUS => mem::size_of::<MpcBus>(),
+            mpspec::MP_IOAPIC => mem::size_of::<MpcIoapic>(),
+            mpspec::MP_INTSRC => mem::size_of::<MpcIntsrc>(),
+            mpspec::MP_LINTSRC => mem::size_of::<MpcLintsrc>(),
             _ => panic!("unrecognized mpc table entry type: {}", type_),
         }
     }
 
     #[test]
     fn bounds_check() {
```
lauralt, line 356: Should we prefix the tests with `test_`?

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: This is fixed in #188 (all the cfgs are in their right place). Is it okay to leave it as is for the time being?

src/vm-vcpu-ref/src/x86_64/mptable.rs
andreeaflorescu: I don't think it's possible to hit those errors because we're checking at the beginning if the full table is going to fit in memory, and if it doesn't that we exit immediately.

gsserge (approved):

lauralt (approved):

PR #188, opened 2021-11-09
Upstream arm POC

andreeaflorescu:
The code for running vmm-reference on arm is not completed. We decided to go ahead with merging the POC which does a basic setup of a VM so that we can test the vm-vcpu-ref crate which we want to upstream on both platforms.

We already have a list of things that need to be improved in respect to this arm poc. The list is available in the crate (with the label [aarch64](https://github.com/rust-vmm/vmm-reference/issues?q=is%3Aissue+is%3Aopen+label%3Aaarch64). If we discover any other problems, depending on their severity we can either:
- open issues and fix them in subsequent PRs
- fix them in this current POC

Any changes requested during the commit will be addressed in separate commits. Doing rebases & force push is taking too much time at this point. The disadvantage in a history should not be that big.


andreeaflorescu:
Still need to do:
- move to vm-fdt from crates.io

gsserge (dismissed):

gsserge (approved):

alexandruag (approved):

PR #189, opened 2021-11-11
Upstream MSR

andreeaflorescu:
This PR moves the MSR related modules to vm-vcpu-ref, and adds tests and documentation so we can prepare the modules for upstream.


src/vm-vcpu-ref/src/x86_64/msrs.rs
```diff
@@ -0,0 +1,80 @@
...
+        // end of x86_64 specific code
+        msr_entry_default(MSR_IA32_TSC),
+        kvm_msr_entry {
+            index: MSR_IA32_MISC_ENABLE,
+            data: u64::from(MSR_IA32_MISC_ENABLE_FAST_STRING),
+            ..Default::default()
+        },
+    ];
+
+    Msrs::from_entries(&raw_msrs).map_err(|_| Error::CreateMsrs)
```
gsserge, line 60: I wonder if there's any useful error information that we can propagate back to the caller instead of discarding it.

gsserge (approved):

src/vm-vcpu-ref/src/x86_64/msrs.rs
andreeaflorescu: The reason I left it without the inner error is because we would need to take an explicit dependency on vmm-sys-util (right now we have a transient dependency on vmm-sys-util through the fam-wrappers feature of kvm-bindings). The `Msrs::from_entries` can only fail due to capacity issues, and I was thinking that's unlikely to change in the future. I don't mind adding it either.

lauralt (approved):

PR #190, opened 2021-11-12
[vm-vcpu-ref] Upstream interrupts - x86

andreeaflorescu:
These patches are preparing the existing interrupts helpers for publishing. For this I added more tests, documentation, and updated some of the abstractions to make it easier to use.


src/vm-vcpu-ref/src/x86_64/interrupts.rs
```diff
@@ -0,0 +1,185 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+// Portions Copyright 2017 The Chromium OS Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the THIRD-PARTY file.
+#![cfg(target_arch = "x86_64")]
+use kvm_bindings::kvm_lapic_state;
+
+// Yanked from byte_order
+macro_rules! generate_read_fn {
```
gsserge, line 11: Any reason for having these two macros if each is only used once?

src/vm-vcpu-ref/src/x86_64/interrupts.rs
```diff
@@ -55,6 +55,20 @@ pub fn set_klapic_reg(klapic: &mut kvm_lapic_state, reg_offset: usize, value: u3
...
+
+/// Set the APIC delivery mode.
+///
+/// # Arguments
+/// * `klapic`: The corresponding `kvm_lapic_state` for which to set the delivery mode.
+/// * `reg_offset`: The offset that identifies the register for which to set the delivery mode.
+///                 Available options exported by this module are: [APIC_LVT0] and [APIC_LVT1].
+/// * `mode`: The APIC mode to set. Available options are:
+///           [Non Maskable Interrupt - NMI](APIC_MODE_NMI) and
+///           [external interrupt - ExtINT](APIC_MODE_EXTINT).
```
lauralt, line 112: Running `cargo doc --open` seems to raise a few warnings because of the references here.

src/vm-vcpu-ref/src/x86_64/interrupts.rs
```diff
@@ -43,23 +43,38 @@ pub const APIC_LVT1: usize = 0x360;
...
+    // The value that we are setting is a u32, which needs 4 bytes of space.
+    // We're here creating a range that can fit the whole value.
     let range = reg_offset..reg_offset + 4;
-    let mut reg = klapic.regs.get_mut(range).expect("set_klapic_reg range");
-    write_le_i32(&mut reg, value as i32)
+    let mut reg = klapic
+        .regs
+        .get_mut(range)
+        .ok_or(Error::InvalidRegisterOffset)?;
+    write_le_i32(&mut reg, value as i32);
```
lauralt, line 92: Should we write a comment about why the conversion from u32 to i32 is safe? The same applies to `get_klapic_reg`.

src/vm-vcpu-ref/src/x86_64/interrupts.rs
```diff
@@ -132,19 +152,27 @@ mod tests {
...
             Error::InvalidRegisterOffset
         );
         assert_eq!(
             get_klapic_reg(&klapic, offset).unwrap_err(),
             Error::InvalidRegisterOffset
         );
     }
+
+    #[test]
+    fn test_delivery_mode() {
```
lauralt, line 179: nit: should we test `SMI` as well?

src/vm-vcpu-ref/src/x86_64/interrupts.rs
andreeaflorescu: I think the idea was to have them as they were defined in `byte_order`. We can also remove the macro and just have it as a function.

src/vm-vcpu-ref/src/x86_64/interrupts.rs
andreeaflorescu: Done.

src/vm-vcpu-ref/src/x86_64/interrupts.rs
andreeaflorescu: I just actually made everything i32 because that's what we end up using.

src/vm-vcpu-ref/src/x86_64/interrupts.rs
andreeaflorescu: Done.

src/vm-vcpu-ref/src/x86_64/interrupts.rs
```diff
@@ -7,36 +7,22 @@
...
-            assert!($type_size == std::mem::size_of::<$data_type>());
-            let mut array = [0u8; $type_size];
-            for (byte, read) in array.iter_mut().zip(input.iter().cloned()) {
-                *byte = read as u8;
-            }
-            <$data_type>::$endian_type(array)
-        }
-    };
+// Helper function that writes the input (which must have 4 bytes) into an i32.
+fn read_le_i32(input: &[i8]) -> i32 {
```
lauralt, line 11: Should we write some tests for these ones?
I think we could also keep the assert and check that input length = 4 here.

gsserge (dismissed):

lauralt (approved):

gsserge (approved):

PR #191, opened 2021-11-16
RFC: add logging functionality

dodsonmg:
Checking to see if this is the kind of logging implementation you're looking for to address #79.

It uses the `log` crate and implements a simple logger that I mostly copied from cloud-hypervisor:
https://github.com/cloud-hypervisor/cloud-hypervisor/blob/f151a8602c1e8b2b17258aef99abd33140622839/src/main.rs#L74-L117

I left most of the `println!` and `eprintln!` statements in `main.rs` and the `api` crates alone because these provide immediate feedback to the CLI user (e.g., printing a help message) and it didn't seem appropriate to redirect these to `stderr` with a timestamp.

I replaced remaining statements in all the crates with appropriate `info!`, `warn!`, or `error!` messages.  I also replaced usage of the `debug!` macro from the `utils` crate with the `debug!` macro provided by the `log` crate.

The implementation defaults to the `Debug` level, meaning only log entries initiated with the `trace!` macro are filtered out.  Alternately, I could implement a command line argument to set the log level, similar to the cloud-hypervisor implementation.

Also, I would appreciate feedback on how I might go about writing unit tests for the logger module.  The tests I've found elsewhere (e.g., for firecracker) support a much more complex logger implementation.

### DIFF
--- a/src/devices/src/legacy/rtc.rs
+++ b/src/devices/src/legacy/rtc.rs
@@ -6,7 +6,7 @@ use vm_device::bus::MmioAddress;
 use vm_device::MutDeviceMmio;
 use vm_superio::{rtc_pl031::NoEvents, Rtc};
 
-use utils::debug;
+use log::debug;
 
 pub struct RtcWrapper(pub Rtc<NoEvents>);
 
--- a/src/logger/mod.rs
+++ b/src/logger/mod.rs
@@ -1,0 +1,73 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// Copyright 2019 Intel Corporation.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use log::LevelFilter;
+
+use std::sync::Mutex;
+
+pub(crate) struct Logger {
+    output: Mutex<Box<dyn std::io::Write + Send>>,
+    start: std::time::Instant,
+}
+
+impl Logger {
+    pub(crate) fn init() {
+        // Default to Debug.
+        // TODO: Accept a command line argument to set the log level.
+        let log_level = LevelFilter::Debug;
+
+        let log_file: Box<dyn std::io::Write + Send> = Box::new(std::io::stderr());
+
+        match log::set_boxed_logger(Box::new(Logger {
+            output: Mutex::new(log_file),
+            start: std::time::Instant::now(),
+        })) {
+            Ok(_) => log::set_max_level(log_level),
+            Err(e) => eprintln!("Failed to initialize logger: {}.", e),
+        }
+    }
+}
+
+impl log::Log for Logger {
+    fn enabled(&self, _metadata: &log::Metadata) -> bool {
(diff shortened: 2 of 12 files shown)


src/devices/src/legacy/rtc.rs
```diff
@@ -6,7 +6,7 @@ use vm_device::bus::MmioAddress;
 use vm_device::MutDeviceMmio;
 use vm_superio::{rtc_pl031::NoEvents, Rtc};
 
-use utils::debug;
```
andreeaflorescu, line 9: The debug macro was doing something special, is this one working the same way? Specifically, we don't want to logs to be filled with data coming from the guest when we have release builds because that can be used to generate a DoS when the log file is not properly rate limited.

src/logger/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+pub(crate) struct Logger {
+    output: Mutex<Box<dyn std::io::Write + Send>>,
+    start: std::time::Instant,
+}
+
+impl Logger {
+    pub(crate) fn init() {
+        // Default to Debug.
+        // TODO: Accept a command line argument to set the log level.
+        let log_level = LevelFilter::Debug;
```
andreeaflorescu, line 18: We can have this as a parameter for the init function, and have a default for it (which can be LevelFilter::Debug). This would make it easier to extend it.

src/logger/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+        let log_file: Box<dyn std::io::Write + Send> = Box::new(std::io::stderr());
```
andreeaflorescu, line 20: This can also be a parameter for which we have a default.

src/vm-vcpu/Cargo.toml
```diff
@@ -17,6 +17,8 @@ utils = { path = "../utils" }
 vm-vcpu-ref = { path = "../vm-vcpu-ref" }
 arch = { path = "../arch" }
 
+log = { version = ">=0.4", features = ["std"] }
```
andreeaflorescu, line 20: Why is the `std` feature required?

andreeaflorescu:
Once we add a configuration for the logger we might find a few things worth testing. Until that point I am not sure we can do anything to test the logger that would not be a "coverage test" (i.e. a test that is not checking anything useful, and it's written just so that the coverage does not drop).

src/logger/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        let now = std::time::Instant::now();
+        let duration = now.duration_since(self.start);
+
+        if record.file().is_some() && record.line().is_some() {
+            writeln!(
+                *(*(self.output.lock().unwrap())),
+                "vmm-reference: {:?}: <{}> {}:{}:{} -- {}",
+                duration,
```
gsserge, line 51: Why duration instead of a wall clock time stamp?

andreeaflorescu:
Closing this PR as it is stale. @dodsonmg feel free to re-open if you want to continue working on it.

PR #192, opened 2021-11-17
Upstream interrupt controller setup for arm

andreeaflorescu:
This PR adds the first arm specific helper in vm-vcpu-ref. This is used for setting up the GIC.
There are a few short comings to the way things are implemented right now because I did not find a way to export the registers and lengths that we're setting in the GIC. This requires some duplicated defines in the FDT setup. More details about it in the commit message.


gsserge (dismissed):

gsserge (dismissed):

src/vm-vcpu-ref/src/aarch64/interrupts.rs
```diff
@@ -0,0 +1,380 @@
...
+            flags: 0,
+        };
+
+        self.device_fd
+            .set_device_attr(&cpu_if_attr)
+            .map_err(|e| Error::SetAttr("cpu", e))
+    }
+
+    fn set_redist_attr(&mut self) -> Result<()> {
+        let redist_addr: u64 =
```
lauralt, line 229: We should add a comment here on why it is safe to use unchecked arithmetic.

src/vm-vcpu-ref/src/aarch64/interrupts.rs
```diff
@@ -0,0 +1,380 @@
...
+    ///
+    /// let gic_config = GicConfig {
+    ///     num_cpus: 1,
+    ///     ..Default::default()
+    /// };
+    ///
+    /// let gic = Gic::new(gic_config, &vm).unwrap();
+    /// let device_fd = gic.device_fd();
+    /// ```
+    pub fn new(config: GicConfig, vm_fd: &VmFd) -> Result<Gic> {
```
lauralt, line 138: I think it would be worth having an Arguments section as well for this one.

lauralt (dismissed):

gsserge (approved):

lauralt (approved):

gsserge (approved):

PR #197, opened 2021-12-08
Cleanup of the Vm & Vcpu interfaces (adding save/restore high level interface)

andreeaflorescu:
The point of this cleanup is to make it easier to work in parallel on extending the vm-vcpu interface with support  for save/restore on both platforms (aarch64 & x86_64), but also to work on upstreaming the `KvmVm` and `KvmVcpu` interfaces as part of vm-vcpu-ref.

The focus of this PR is not on having a complete implementation for save/restore, but to provide a high level design for it such that the work can be easily split among more people. As such, the save/restore functionality is not tested, this will be added as a follow up PR (coming after the vm-vcpu upstreaming).

### DIFF
--- a/src/vm-vcpu-ref/src/x86_64/msrs.rs
+++ b/src/vm-vcpu-ref/src/x86_64/msrs.rs
@@ -1,21 +1,37 @@
 // Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 use kvm_bindings::{kvm_msr_entry, Msrs};
+use kvm_ioctls::Kvm;
 
-use crate::x86_64::msr_index::{
-    MSR_CSTAR, MSR_IA32_MISC_ENABLE, MSR_IA32_MISC_ENABLE_FAST_STRING, MSR_IA32_SYSENTER_CS,
-    MSR_IA32_SYSENTER_EIP, MSR_IA32_SYSENTER_ESP, MSR_IA32_TSC, MSR_KERNEL_GS_BASE, MSR_LSTAR,
-    MSR_STAR, MSR_SYSCALL_MASK,
-};
+use crate::x86_64::msr_index::*;
 
 /// Errors associated with operations on MSRs.
 #[derive(Debug, PartialEq)]
 pub enum Error {
     /// Failed to initialize MSRS.
     CreateMsrs,
+    /// Failed to get supported MSRs.
+    GetSupportedMSR(kvm_ioctls::Error),
 }
 /// Specialized result type for operations on MSRs.
 pub type Result<T> = std::result::Result<T, Error>;
+
+/// Base MSR for APIC
+const APIC_BASE_MSR: u32 = 0x800;
+
+/// Number of APIC MSR indexes
+const APIC_MSR_INDEXES: u32 = 0x400;
+
+/// Custom MSRs fall in the range 0x4b564d00-0x4b564dff
+const MSR_KVM_WALL_CLOCK_NEW: u32 = 0x4b56_4d00;
+const MSR_KVM_SYSTEM_TIME_NEW: u32 = 0x4b56_4d01;
+const MSR_KVM_ASYNC_PF_EN: u32 = 0x4b56_4d02;
+const MSR_KVM_STEAL_TIME: u32 = 0x4b56_4d03;
+const MSR_KVM_PV_EOI_EN: u32 = 0x4b56_4d04;
+
+/// Taken from arch/x86/include/asm/msr-index.h
+const MSR_IA32_SPEC_CTRL: u32 = 0x0000_0048;
+const MSR_IA32_PRED_CMD: u32 = 0x0000_0049;
 
 /// Creates and populates required MSR entries for booting Linux on X86_64.
 ///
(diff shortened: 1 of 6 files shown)


src/vm-vcpu/src/vm.rs
```diff
@@ -667,7 +650,7 @@ mod tests {
         let mut guest_memory = default_memory();
 
         let mut vm = create_vm_and_vcpus(num_vcpus, &mut guest_memory);
-
+        println!("{:#?}", vm.vcpus[0].vcpu_fd.get_regs().unwrap());
```
andreeaflorescu, line 653: To delete.

src/vm-vcpu/src/vm.rs
```diff
@@ -678,9 +661,8 @@ mod tests {
         assert_eq!(vm_state.pic_slave.chip_id, KVM_IRQCHIP_PIC_SLAVE);
         assert_eq!(vm_state.ioapic.chip_id, KVM_IRQCHIP_IOAPIC);
 
-        for vcpu in vm.vcpus.iter_mut() {
-            // nothing really special about the state that we can check.
-            vcpu.save_state().unwrap();
-        }
+        // At this point the vcpus have not been running, so the REGS should
```
andreeaflorescu, line 664: Needs update.

src/vm-vcpu-ref/src/x86_64/msrs.rs
```diff
@@ -1,22 +1,55 @@
 // Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
 use kvm_bindings::{kvm_msr_entry, Msrs};
+use kvm_ioctls::Kvm;
 
 use crate::x86_64::msr_index::{
-    MSR_CSTAR, MSR_IA32_MISC_ENABLE, MSR_IA32_MISC_ENABLE_FAST_STRING, MSR_IA32_SYSENTER_CS,
-    MSR_IA32_SYSENTER_EIP, MSR_IA32_SYSENTER_ESP, MSR_IA32_TSC, MSR_KERNEL_GS_BASE, MSR_LSTAR,
-    MSR_STAR, MSR_SYSCALL_MASK,
+    MSR_MTRRcap, MSR_CONFIG_TDP_CONTROL, MSR_CONFIG_TDP_LEVEL_1, MSR_CONFIG_TDP_LEVEL_2,
```
gsserge, line 7: This is a huge import tbh. I wouldn't mind `use crate::x86_64::msr_index::*`.

src/vm-vcpu-ref/src/x86_64/msrs.rs
```diff
@@ -60,9 +93,177 @@ pub fn create_boot_msr_entries() -> Result<Msrs> {
...
+    SINGLE_MSR!(MSR_TURBO_RATIO_LIMIT1),
+    SINGLE_MSR!(MSR_IA32_DEBUGCTLMSR),
+    SINGLE_MSR!(MSR_IA32_LASTBRANCHFROMIP),
+    SINGLE_MSR!(MSR_IA32_LASTBRANCHTOIP),
+    SINGLE_MSR!(MSR_IA32_LASTINTFROMIP),
+    SINGLE_MSR!(MSR_IA32_LASTINTTOIP),
+    SINGLE_MSR!(MSR_IA32_POWER_CTL),
+    MSR_RANGE!(
+        // IA32_MTRR_PHYSBASE0
+        0x200, 0x100
```
gsserge, line 175: Just to make sure, are there 256 registers in this range?

src/vm-vcpu-ref/src/x86_64/msrs.rs
```diff
@@ -60,9 +93,177 @@ pub fn create_boot_msr_entries() -> Result<Msrs> {
...
+    let mut msrs =
+        Msrs::new(msr_list.as_fam_struct_ref().nmsrs as usize).map_err(|_| Error::CreateMsrs)?;
+    let indices = msr_list.as_slice();
+    let msr_entries = msrs.as_mut_slice();
+    // We created the msrs from the msr_list. If the size is not the same,
+    // there is a fatal programming error.
+    assert_eq!(indices.len(), msr_entries.len());
+    for (pos, index) in indices.iter().enumerate() {
+        msr_entries[pos].index = *index;
+    }
```
gsserge, line 259: Too bad that the logic of constructing `Msrs` cannot go into the constructor itself due to bindgen, and instead leaks into the client code :(

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -187,6 +316,67 @@ impl KvmVcpu {
         Ok(vcpu)
     }
 
+    #[cfg(target_arch = "x86_64")]
+    fn set_state(&mut self, state: VcpuState) -> Result<()> {
```
gsserge, line 320: I wonder if it's safer to have this method consume self instead of mutating it in place. At the moment it's only used in `from_state()` below where the usage is def safe because a new `KvmVcpu` object is created locally and either returned being correct or dropped in case of an error - a malformed object is never returned.

However, the method in its current form does not prevent setting an existing object to incorrect state and keep using it if the client code is not careful.

src/vm-vcpu/src/vm.rs
```diff
@@ -113,36 +167,97 @@ impl Default for VmRunState {
...
         #[cfg(target_arch = "x86_64")]
         vm.setup_irq_controller()?;
 
+        vm.create_vcpus(bus, vcpus_config, guest_memory)?;
+
+        Ok(vm)
+    }
+
+    #[cfg(target_arch = "x86_64")]
+    fn set_state(&mut self, state: VmState) -> Result<()> {
```
gsserge, line 219: Similar to `KvmVcpu` https://github.com/rust-vmm/vmm-reference/pull/197/files#r766628831

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -406,18 +547,25 @@ impl KvmVcpu {
...
-                .set_one_reg(reg_id, data)
-                .map_err(Error::SetReg)?;
+    pub fn run(&mut self, instruction_pointer: Option<GuestAddress>) -> Result<()> {
+        if let Some(ip) = instruction_pointer {
+            #[cfg(target_arch = "x86_64")]
+            self.configure_regs(ip)?;
+            #[cfg(target_arch = "aarch64")]
+            if self.config.id == 0 {
+                let data = ip.0;
+                println!("data={}", data);
```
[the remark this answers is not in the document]
andreeaflorescu: No, will delete.

src/vm-vcpu/src/vm.rs
```diff
@@ -316,7 +323,7 @@ impl<EH: 'static + ExitHandler + Send> KvmVm<EH> {
```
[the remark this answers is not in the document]
andreeaflorescu: Yes, this one is still accurate. The one about reusing the GIC was deleted because it was a leftover from before we merged the arm POC branch.

src/vmm/src/lib.rs
```diff
@@ -595,47 +590,6 @@ impl Vmm {
         Ok(kernel_load_addr)
     }
 
-    // Create guest vCPUs based on the passed vCPU configurations.
-    fn create_vcpus(&mut self, vcpu_cfg: &VcpuConfig) -> Result<()> {
```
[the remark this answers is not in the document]
andreeaflorescu: Yap, I'll disambiguate the commit message.

src/vmm/src/lib.rs
```diff
@@ -996,25 +952,30 @@ mod tests {
...
 
+        // Creating 254 also works (that's the maximum number on x86 when using MP Table).
+        vmm_config.vcpu_config = VcpuConfig { num: 254 };
         {
-            let vmm_config = default_vmm_config();
-            let mut vmm = mock_vmm(vmm_config);
-            // Creating 255 also works.
-            assert!(vmm.create_vcpus(&VcpuConfig { num: u8::MAX }).is_ok());
+            Vmm::try_from(vmm_config.clone()).unwrap();
+            assert!(Vmm::try_from(vmm_config.clone()).is_ok());
```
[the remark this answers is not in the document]
andreeaflorescu: It was failing in the tests, and I did not know why. I think I'll just keep the `unwrap` in the test because when things don't work we have a better idea of what actually failed.

src/vm-vcpu/src/vcpu/mod.rs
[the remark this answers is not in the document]
andreeaflorescu: Apparently that's how we merged the arm POC, but I'll still delete this println, it will just be an unrelated code change to this PR as it's not introduced here.

src/vm-vcpu/src/vcpu/mod.rs
andreeaflorescu: Since this is an internal function, I think I would rather add a comment instead and say that the errors coming from this function must be immediately return because otherwise you end up in an undefined state. Does that sound right to you?

src/vmm/src/lib.rs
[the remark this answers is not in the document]
andreeaflorescu: Done. We're now just keeping the unwrap.

gsserge (approved):

lauralt (approved):

PR #211, opened 2022-03-21
fix testcase test_run_reference_vmm.py::test_reference_vmm_num_vcpus

iscas-glm:
in test_reference_vmm_num_vcpus testcase,it calls real check function expect_vcpus(vmm_process, expected_vcpus)
which extracts "siblings" in /proc/cpuinfo file,while the number of siblings on a processor is the total number of execution units within that processor,not the meaning of real physical vcpus.

Early I started by trying to send the following command from the test function to the child process that started the VM

`cat /proc/cpuinfo| grep "physical id"|wc -l`

But it seems that the vmm_process.stdin.write() function can only write 32 characters to the terminal, and I don't know why.

I ended up getting the number of virtual machine instance cpus in a different way.

### DIFF
--- a/tests/test_run_reference_vmm.py
+++ b/tests/test_run_reference_vmm.py
@@ -225,16 +225,37 @@ def expect_vcpus(vmm_process, expected_vcpus):
     # dependent on it, so it's just fine to call it again, less than ideal, but not
     # wrong.
     setup_stdout_nonblocking(vmm_process)
+    giveup_after = timeout
+    then = time.time()
+    cnt=0
+    while True:
+        try:
+            data = os.read(vmm_process.stdout.fileno(), 4096)
+            output_lines = data.split(b'\r\n')
+            for line in output_lines:
+                if expected_string in line.decode():
+                    cnt+=1
+            return cnt
+        except BlockingIOError as _:
+            # Raised on `EWOULDBLOCK`, so it's better to wait for sometime before retrying
+            time.sleep(1)
+            now = time.time()
+            if now - then > giveup_after:
+                raise TimeoutError(
+                    "Timed out {} waiting for {}".format(now - then, expected_string))
+        except Exception as _:
+            raise
+       
+    return cnt
 
-    # /proc/cpuinfo displays info about each vCPU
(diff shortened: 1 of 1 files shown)


tests/test_run_reference_vmm.py
```diff
@@ -225,16 +225,37 @@ def expect_vcpus(vmm_process, expected_vcpus):
...
-    # /proc/cpuinfo displays info about each vCPU
-    vmm_process.stdin.write(b'cat /proc/cpuinfo\n')
-    vmm_process.stdin.flush()
-
-    siblings_line = expect_string(vmm_process, "siblings")
-    actual_vcpus = int(siblings_line.split(":")[1].strip())
 
-    assert actual_vcpus == expected_vcpus, \
-        "Expected {}, found {} vCPUs".format(expected_vcpus, actual_vcpus)
+def expect_vcpus(vmm_process, expected_vcpus):
```
andreeaflorescu, line 252: Why do we have another function named expect_vcpus?

tests/test_run_reference_vmm.py
```diff
@@ -225,16 +225,37 @@ def expect_vcpus(vmm_process, expected_vcpus):
     # dependent on it, so it's just fine to call it again, less than ideal, but not
     # wrong.
     setup_stdout_nonblocking(vmm_process)
+    giveup_after = timeout
+    then = time.time()
+    cnt=0
+    while True:
```
andreeaflorescu, line 231: I don't understand why this change is needed. Can you please explain it? Also, it should be part of a separate commit.

andreeaflorescu: If we need something more complex than just a `vmm_process.stdin.write(b'cmd')` then we should have this functionality in a separate function, and then call it in all the other functions as well. It's strange to have different ways of sending commands to the VMM depending on the test function.

andreeaflorescu (changes requested):
Can you also update the commit message as explained in the failed text?

andreeaflorescu:
Closing this due to lack of activity. Feel free to re-open it if you want to continue to work on it.

Issue #126, opened 2021-05-25
Create a minimal FDT


andreeaflorescu:
Import the rust-vmm/vm-fdt code and define a minimal FDT. At a minimum the FDT should at least have the memory, vcpu, interrupts, and timer nodes.The FDT then needs to be initialized as part of the Vmm crate. We need to find the best place for this initialization.

uran0sH:
Could I do it if no one work on it?

andreeaflorescu:
Yes, you can pick this up, but it might not be the best good first issue. The minimal FDT is already created, but there are some shortcomings to it, namely it's pretty hardcoded and the initialization is happening here: https://github.com/rust-vmm/vmm-reference/blob/ceacce791beed583dcc6cd6df95e72c47ef0f3d4/src/vmm/src/lib.rs#L314. One of the problems is that we need the vcpu count, and we did a pretty ugly hack to get it, as briefly explained here: https://github.com/rust-vmm/vmm-reference/blob/ceacce791beed583dcc6cd6df95e72c47ef0f3d4/src/vmm/src/lib.rs#L184

andreeaflorescu:
And here is the code for setting up the FDT: https://github.com/rust-vmm/vmm-reference/blob/ceacce791beed583dcc6cd6df95e72c47ef0f3d4/src/arch/src/lib.rs#L66

andreeaflorescu:
We can reduce the scope of this issue so that you can work on it. Instead of fixing all the shortcomings of this, we can do the following instead:
- remove the hard coded addresses for the RTC and the Serial Console device, and pass them as a parameter instead (https://github.com/rust-vmm/vmm-reference/blob/ceacce791beed583dcc6cd6df95e72c47ef0f3d4/src/arch/src/lib.rs#L177 and https://github.com/rust-vmm/vmm-reference/blob/ceacce791beed583dcc6cd6df95e72c47ef0f3d4/src/arch/src/lib.rs#L247)
- as this will add quite a few parameters to the `create_fdt` function, we can re-implement it using the Builder Pattern: https://doc.rust-lang.org/1.0.0/style/ownership/builders.html

The second bullet point can be done afterwards.

uran0sH:
Ok, I also have send message to you in Slack.

PR #212, opened 2022-03-24
Remove hard coded addresses for the RTC and Serial

uran0sH:
### Summary of the PR

We need this to make devices add dynamically to the FDT.

issue: #126 
### Requirements

requirements:

  description line.

### DIFF
--- a/src/arch/src/lib.rs
+++ b/src/arch/src/lib.rs
@@ -68,6 +68,8 @@ pub fn create_fdt<T: GuestMemory>(
     num_vcpus: u32,
     guest_mem: &T,
     fdt_load_offset: u64,
+    serial_addr: u64,
+    rtc_addr: u64,
 ) -> Result<()> {
     let mut fdt = FdtWriter::new()?;
 
@@ -85,8 +87,8 @@ pub fn create_fdt<T: GuestMemory>(
     create_memory_node(&mut fdt, mem_size)?;
     create_cpu_nodes(&mut fdt, num_vcpus)?;
     create_gic_node(&mut fdt, true, num_vcpus as u64)?;
-    create_serial_node(&mut fdt)?;
-    create_rtc_node(&mut fdt)?;
+    create_serial_node(&mut fdt, serial_addr)?;
+    create_rtc_node(&mut fdt, rtc_addr)?;
     create_timer_node(&mut fdt, num_vcpus)?;
     create_psci_node(&mut fdt)?;
     create_pmu_node(&mut fdt, num_vcpus)?;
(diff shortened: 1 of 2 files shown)


andreeaflorescu:
LGTM, I don't think we need to add a test for this one at the moment because we have some subsequent changes that are needed. Can you remove the commented lines? Also, it would be nice for the commit message to respond to the question of "why do we need this change?". The answer should be something along the lines of: without this it is not possible to dynamically add devices to the FDT.

andreeaflorescu (dismissed):

uran0sH:
I use the Builder Pattern to re-implement `create_fdt`. plz review.

src/arch/src/lib.rs
```diff
@@ -63,43 +63,76 @@ impl From<GuestMemoryError> for Error {
...
+    T: GuestMemory,
+{
+    pub fn new(cmdline: &'a str, num_vcpus: u32, guest_mem: &'a T, fdt_load_offset: u64) -> Self {
+        FdtBuilder {
+            cmdline,
+            num_vcpus,
+            guest_mem,
+            fdt_load_offset,
+            serial_addr: 0x40000000,
+            rtc_addr: 0x40001000,
```
andreeaflorescu, line 86: These should be set by the user. We shouldn't have defaults for them because there is no good default value here.

src/arch/src/lib.rs
```diff
@@ -63,43 +63,76 @@ impl From<GuestMemoryError> for Error {
 
 pub type Result<T> = std::result::Result<T, Error>;
 
-pub fn create_fdt<T: GuestMemory>(
-    cmdline: &str,
+pub struct FdtBuilder<'a, T: GuestMemory> {
+    cmdline: &'a str,
     num_vcpus: u32,
-    guest_mem: &T,
+    guest_mem: &'a T,
```
andreeaflorescu, line 69: Why do both guest_mem and cmdline use the same lifetime?

src/arch/src/lib.rs
```diff
@@ -63,43 +63,76 @@ impl From<GuestMemoryError> for Error {
...
-    let fdt_address = GuestAddress(AARCH64_PHYS_MEM_START + fdt_load_offset);
-    guest_mem.write_slice(fdt_final.as_slice(), fdt_address)?;
+}
 
-    Ok(())
+impl<'a, T> FdtBuilder<'a, T>
+where
+    T: GuestMemory,
+{
+    pub fn new(cmdline: &'a str, num_vcpus: u32, guest_mem: &'a T, fdt_load_offset: u64) -> Self {
```
andreeaflorescu, line 79: The builder does not need the GuestMemory object. Only when we actually get an FDT we need the GuestMemory so we can write it to memory. We need to do here some separation of concerns:
- building the FDT -> through the FdtBuilder
- operating the Fdt -> possibly it's better to create an Fdt object.
I need to think a bit what is the better approach here, I'll get back to you, but feel free to experiment with this yourself.

lauralt: Since this requires more work, I think it would be better to split the PR in the two commits. First commit looks good to me, and we can merge it.

lauralt (approved):

andreeaflorescu (approved):

PR #214, opened 2022-03-29
Reimplement create_fdt using the Builder Pattern

uran0sH:
### Summary of the PR

In the future, we might add more parameters to the create_fdt.
So it's convenient for us to add more parameters.

issue: #126 

### Requirements

requirements:

  description line.
test.

### DIFF
--- a/src/arch/src/lib.rs
+++ b/src/arch/src/lib.rs
@@ -3,9 +3,7 @@
 // found in the LICENSE file.
 
 pub use vm_fdt::{Error as FdtError, FdtWriter};
-use vm_memory::{
-    guest_memory::Error as GuestMemoryError, Bytes, GuestAddress, GuestMemory, GuestMemoryRegion,
-};
+use vm_memory::{guest_memory::Error as GuestMemoryError, Bytes, GuestAddress, GuestMemory};
 // This is an arbitrary number to specify the node for the GIC.
 // If we had a more complex interrupt architecture, then we'd need an enum for
 // these.
@@ -47,6 +45,7 @@ const AARCH64_PMU_IRQ: u32 = 7;
 pub enum Error {
     Fdt(FdtError),
     Memory(GuestMemoryError),
+    MissingRequiredConfig(String),
 }
 
 impl From<FdtError> for Error {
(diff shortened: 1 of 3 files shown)


src/arch/src/lib.rs
```diff
@@ -63,43 +61,88 @@ impl From<GuestMemoryError> for Error {
...
+
+        // The whole thing is put into one giant node with s
+        // ome top level properties
+        let root_node = fdt.begin_node("")?;
+        fdt.property_u32("interrupt-parent", PHANDLE_GIC)?;
+        fdt.property_string("compatible", "linux,dummy-virt")?;
+        fdt.property_u32("#address-cells", 0x2)?;
+        fdt.property_u32("#size-cells", 0x2)?;
+
+        create_chosen_node(&mut fdt, self.cmdline.unwrap_or_default())?;
```
andreeaflorescu, line 114: The cmdline, num_vcpus, and mem_size should be mandatory. if these are not specified, than we should return an error.

src/arch/src/lib.rs
```diff
@@ -63,43 +61,88 @@ impl From<GuestMemoryError> for Error {
...
+pub struct FdtBuilder<'a> {
+    cmdline: Option<&'a str>,
+    mem_size: Option<u64>,
+    num_vcpus: Option<u32>,
+    serial_console: Option<(u64, u64)>,
+    rtc: Option<(u64, u64)>,
+}
 
-    Ok(())
+impl<'a> FdtBuilder<'a> {
```
andreeaflorescu, line 73: We should be able to add some unit tests now that the code is structured this way. 

What we can check is at least the error cases (once we add them):
- creating an fdt without either of num_vcpus, mem, cmdline returns an error
- writing the fdt outside the reserved memory fails
We can also test that a "good FDT" with everything we need in it can be successfully built from the builder.

I am still thinking about a better way to test the internals of the FDT, but that might actually not be possible at this moment (i.e. checking somehow that the RTC is actually added to the FDT blob if the `with_rtc` function is called).

andreeaflorescu (commented):
I just have one more comment, otherwise LGTM. Great progress on the PR!

andreeaflorescu (dismissed):

src/arch/src/lib.rs
lauralt: If these are mandatory, should we mark them as such in the FdtBuilder? Looks a bit unusual to have all the struct members as optional.

uran0sH: If we don't use Option, we need to set a default value to them. And there doesn't seem to be a suitable default value to set to them

lauralt: I was thinking to pass the required parameters to `new`, similar to how it is described [here](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html).

src/vmm/src/lib.rs
```diff
@@ -1043,4 +1043,44 @@ mod tests {
...
+            let fdt = FdtBuilder::new()
+                .with_cmdline(vmm.kernel_cfg.cmdline.as_str())
+                .with_num_vcpus(vmm.num_vcpus.try_into().unwrap())
+                .with_mem_size(mem_size)
+                .with_serial_console(0x40000000, 0x1000)
+                .with_rtc(0x40001000, 0x1000)
+                .create_fdt()
+                .unwrap();
+            assert!(fdt.write_to_mem(&vmm.guest_memory, fdt_offset).is_ok());
+        }
```
lauralt, line 1070: Hmm, what is the reason for adding this test case? Looks like it simply copies the code from `setup_fdt`, and you already check above that `setup_fdt` works as expected.

uran0sH: Yes, this is a duplicate, I will delete it

src/arch/src/lib.rs
uran0sH: What do you think? @andreeaflorescu

andreeaflorescu: I think it's better to not pass them in new because then the new function will have too many parameters. I prefer to keep them as they are now. It is typical for the builder to have all parameters as Optional because that's how you get to pass them one by one. Parameters that are required can only be verified at the end, when the object we want to build is done with the configuration.

andreeaflorescu: When you only have one parameter that's not optional, then it's fine to pass it to the `new` function, especially if it's something obvious.

lauralt: :+1:

lauralt (approved):

andreeaflorescu (approved):

Issue #198, opened 2021-12-13
consume i8042 device from vm-superio


lauralt:
At this point we don't have an implementation of the i8042 device in vmm-reference, and we don't do anything for [the related vcpu exits](https://github.com/rust-vmm/vmm-reference/blob/ceacce791beed583dcc6cd6df95e72c47ef0f3d4/src/vm-vcpu/src/vcpu/mod.rs#L685). We should consume the device from [vm-superio](https://github.com/rust-vmm/vm-superio/blob/main/crates/vm-superio/src/i8042.rs) in order to gracefully handle the shutdown.

andreeaflorescu:
Task breakdown:
- Define an i8042Wrapper in devices/src/legacy that extends the upstream i8042. This should be defined similar to the RtcWrapper or the SerialWrapper.
- Implement the `MutDevicePio` trait for the i8042 Wrapper. A good example of how this can be done is here: https://github.com/google/crosvm/blob/cbddc592dd256155f14054e10b5c539c3c6c794a/devices/src/i8042.rs#L23
- in vmm/src/lib.rs initialize the i8042 wrapper and add it to the `device_mgr` (this should also include calling register_irqfd)

Once these things are done, we can test the integration works by removing "reboot=t" from the command line defined in `vmm/src/lib.rs` and check that we still properly run the tests that expect the vmm to exit.

lauralt:
Fixed by #215

PR #215, opened 2022-03-30
consume i8042 device from vm-superio

ramyak-mehra:
Signed-off-by: Ramyak Mehra <rmehra_be19@thapar.edu>

### Summary of the PR
* Added I8042 wrapper inside devices/legacy.
* add_i8042_deive func inside src/vmm/lib.rs on the vmm struct.
* handled VcpuExit::IoOut at address i8042 port, inside vm-vcpu/src/vcpu/mod.rs

issue: #198 

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/devices/src/legacy/i8042.rs
+++ b/src/devices/src/legacy/i8042.rs
@@ -1,0 +1,42 @@
+// Copyright 2022 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+use std::convert::TryInto;
+use vm_device::bus::{PioAddress, PioAddressOffset};
+use vm_device::MutDevicePio;
+use vm_superio::I8042Device;
+
+use utils::debug;
+
+use super::EventFdTrigger;
+pub struct I8042Wrapper(pub I8042Device<EventFdTrigger>);
+
+impl MutDevicePio for I8042Wrapper {
+    fn pio_read(&mut self, _base: PioAddress, offset: PioAddressOffset, data: &mut [u8]) {
+        if data.len() != 1 {
+            debug!("Invalid I8042 data length on read: {}", data.len());
+            return;
+        }
+        match offset.try_into() {
+            Ok(offset) => {
+                self.0.read(offset);
+            }
+            Err(_) => debug!("Invalid I8042 read offset."),
+        }
+    }
+
+    fn pio_write(&mut self, _base: PioAddress, offset: PioAddressOffset, data: &[u8]) {
+        if data.len() != 1 {
+            debug!("Invalid I8042 data length on write: {}", data.len());
+            return;
+        }
+        match offset.try_into() {
+            Ok(offset) => {
+                if self.0.write(offset, data[0]).is_err() {
+                    debug!("Failed to write to I8042.");
+                }
+            }
+            Err(_) => debug!("Invalid I8042 write offset"),
+        }
+    }
+}
--- a/src/vm-vcpu/src/vcpu/mod.rs
+++ b/src/vm-vcpu/src/vcpu/mod.rs
@@ -684,7 +684,18 @@ impl KvmVcpu {
                                 }
                             } else if addr == 0x060 || addr == 0x061 || addr == 0x064 {
                                 // Write at the i8042 port.
+                                //i8042 is registered at port 0x64.
                                 // See https://wiki.osdev.org/%228042%22_PS/2_Controller#PS.2F2_Controller_IO_Ports
+                                #[cfg(target_arch = "x86_64")]
+                                if self
+                                    .device_mgr
+                                    .lock()
+                                    .unwrap()
+                                    .pio_write(PioAddress(addr), data)
+                                    .is_err()
+                                {
+                                    debug!("Failed to write to i8042 port")
+                                }
                             } else if (0x070..=0x07f).contains(&addr) {
                                 // Write at the RTC port.
                             } else {
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -45,6 +45,8 @@ use vm_device::device_manager::PioManager;
 #[cfg(target_arch = "aarch64")]
 use vm_memory::GuestMemoryRegion;
 use vm_memory::{GuestAddress, GuestMemory, GuestMemoryMmap};
+#[cfg(target_arch = "x86_64")]
+use vm_superio::I8042Device;
 #[cfg(target_arch = "aarch64")]
 use vm_superio::Rtc;
 use vm_superio::Serial;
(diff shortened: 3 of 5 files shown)


ramyak-mehra:
I see the PIO bus is only available on x86_64. Should arch use MMIO or we are to leave the implementation of it for now?

andreeaflorescu:
> I see the PIO bus is only available on x86_64. Should arch use MMIO or we are to leave the implementation of it for now?

The i8042 is only used on x86_64. You should make sure to mark it as such using platform configuration macros.

ramyak-mehra:
> > I see the PIO bus is only available on x86_64. Should arch use MMIO or we are to leave the implementation of it for now?
> 
> The i8042 is only used on x86_64. You should make sure to mark it as such using platform configuration macros.

I fixed the platform configuration issues. Is there a way we can test this implementation.

src/vm-vcpu/src/vcpu/mod.rs
```diff
@@ -684,7 +684,17 @@ impl KvmVcpu {
                                 }
                             } else if addr == 0x060 || addr == 0x061 || addr == 0x064 {
                                 // Write at the i8042 port.
+                                //i8042 is registered at port 0x64.
                                 // See https://wiki.osdev.org/%228042%22_PS/2_Controller#PS.2F2_Controller_IO_Ports
+                                if self
```
lauralt, line 690: This is not exactly correct. Even though for the writes that result in cpu exit, it is expected that `addr` will be 0x64, we shouldn't make any assumptions at this point about the exact value. A driver can also write to 0x60, see example [here](https://github.com/firecracker-microvm/firecracker/blob/main/src/devices/src/legacy/i8042.rs#L290), but we don't offer this support in vm-superio.
It is enough to just use `addr` instead of the hardcoded value below, and the i8042 logic will know how to handle different addresses (at this point it does nothing for addresses != 0x64).

src/devices/src/legacy/i8042.rs
```diff
@@ -0,0 +1,38 @@
...
+const CMD_RESET_CPU: u8 = 0xFE;
+
+pub struct I8042Wrapper(pub I8042Device<EventFdTrigger>);
+
+impl MutDevicePio for I8042Wrapper {
+    fn pio_read(&mut self, _base: PioAddress, offset: PioAddressOffset, data: &mut [u8]) {
+        match offset.try_into() {
+            Ok(offset) => {
+                let read_data = self.0.read(offset);
+                data[0] = read_data;
```
lauralt, line 23: you can assign here `self.0.read(offset)`, and not use the auxiliary variable `read_data`.

src/devices/src/legacy/i8042.rs
```diff
@@ -0,0 +1,38 @@
...
+            Err(_) => debug!("Invalid I8042 read offset."),
+        }
+    }
+
+    fn pio_write(&mut self, _base: PioAddress, _offset: PioAddressOffset, data: &[u8]) {
+        if data.len() != 1 {
+            debug!("Invalid I8042 data length on write: {}", data.len());
+            return;
+        }
+        if let Err(_) = self.0.write(COMMAND_OFFSET, CMD_RESET_CPU) {
```
lauralt: By doing this, you always try to reset the cpu, even though the driver sent another command. Here we should just propagate the `offset` and `data[0]`.

src/devices/src/legacy/i8042.rs
lauralt: We should also check the size of `data`, similarly to what we do for `write`.

src/vmm/src/lib.rs
```diff
@@ -487,6 +488,22 @@ impl Vmm {
...
         Ok(())
     }
+    // Create and add a i8042 device to the VMM.
+    fn add_i8042_device(&mut self) -> Result<()> {
+        let reset_evt = EventFdTrigger::new(libc::EFD_NONBLOCK).map_err(Error::IO)?;
+        let i8042_device = Arc::new(Mutex::new(I8042Wrapper(I8042Device::new(
+            reset_evt.try_clone().map_err(Error::IO)?,
+        ))));
+        self.vm.register_irqfd(&reset_evt, 1)?;
+        let range = PioRange::new(PioAddress(0x064), 0x8).unwrap();
```
lauralt, line 505: The base of a device is where its range starts, i.e. 0x060, and its length should be 0x5 (0x60-0x64). [Here](https://wiki.osdev.org/%228042%22_PS/2_Controller) you can find the specification for i8042.

src/devices/src/legacy/i8042.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2021 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
andreeaflorescu, line 1: this should be 2022.

andreeaflorescu:
> > > > I see the PIO bus is only available on x86_64. Should arch use MMIO or we are to leave the implementation of it for now?
> > > 
> > > 
> > > The i8042 is only used on x86_64. You should make sure to mark it as such using platform configuration macros.
> > 
> > 
> > I fixed the platform configuration issues. Is there a way we can test this implementation?
> 
> Andreea explained [here](https://github.com/rust-vmm/vmm-reference/issues/198#issuecomment-1081542065) at the end how you can test it. [Here](https://github.com/rust-vmm/vmm-reference/blob/main/src/vmm/src/lib.rs#L98) is the command line definition. We will probably also need to remove `i8042.nokbd`.

Can you try to see if this work? @ramyak-mehra

ramyak-mehra:
> > > > > I see the PIO bus is only available on x86_64. Should arch use MMIO or we are to leave the implementation of it for now?
> > > > 
> > > > 
> > > > The i8042 is only used on x86_64. You should make sure to mark it as such using platform configuration macros.
> > > 
> > > 
> > > I fixed the platform configuration issues. Is there a way we can test this implementation?
> > 
> > 
> > Andreea explained [here](https://github.com/rust-vmm/vmm-reference/issues/198#issuecomment-1081542065) at the end how you can test it. [Here](https://github.com/rust-vmm/vmm-reference/blob/main/src/vmm/src/lib.rs#L98) is the command line definition. We will probably also need to remove `i8042.nokbd`.
> 
> Can you try to see if this work? @ramyak-mehra

On my local machine?

andreeaflorescu:
> > Can you try to see if this work? @ramyak-mehra
> 
> On my local machine?

I mean to update the PR to modify the command line and check if the exit in tests still works. Now that's done, so it looks good to me. Can you also fix the failing tests? And maybe squash the commits.

ramyak-mehra:
I did change the default kernel cmd line, but in the python tests we are passing the reboot=t and i8042.nokbd. should I remove it from there as well? I ran the tests by modifying that and the tests passed.

andreeaflorescu:
> I did change the default kernel cmd line, but in the python tests we are passing the reboot=t and i8042.nokbd. should I remove it from there as well? I ran the tests by modifying that and the tests passed.

We should add a test with reboot=t and i8042.nokbd and another one with the default cmdline.

andreeaflorescu (approved):

lauralt (approved):

PR #217, opened 2022-05-17
Init virtio console.

RaduNiculae:
Added a virtio console device to replace the UART serial console.
The implementation is divided between vm-virtio and vmm-reference.

Signed-off-by: Niculae Radu <r.niculae99@gmail.com>

### Summary of the PR

Added a virtio console device with similar queue handling as the block and 
network devices in vmm-reference. Part of the implementation is found in
the vm-virtio repository. For now it is linked to my personal repo until it
will be merged.

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.
- [ ] Any newly added `unsafe` code is properly documented.


lauralt (commented):
Nice work, I'll finish the review tomorrow.

src/devices/src/virtio/console/inorder_handler.rs
```diff
@@ -0,0 +1,97 @@
...
+    pub fn process_transmitq(&mut self) -> result::Result<(), Error>{
+        // To see why this is done in a loop, please look at the `Queue::enable_notification`
+        // comments in `virtio_queue`.
+        loop {
+            self.transmitq.disable_notification()?;
+
+            while let Some(mut chain) = self.transmitq.iter()?.next() {
+                let read_len = self.console.process_transmitq_chain(&mut chain, &mut self.output);
+
+                self.transmitq.add_used(chain.head_index(), read_len as u32)?;
```
lauralt, line 50: Converting from an usize to an u32 is not implicitly safe, we should explain why it is safe or in case it's not, think about how to make things safe.

src/devices/src/virtio/console/inorder_handler.rs
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    pub fn process_receiveq(&mut self) -> result::Result<(), Error>{
+        // To see why this is done in a loop, please look at the `Queue::enable_notification`
+        // comments in `virtio_queue`.
+
+        let mut notify = false;
+
+        loop {
+            if self.console.is_input_buffer_empty() {
```
lauralt, line 70: Hmm I think we are already checking this thing in the upstream function, do we need to do this in two places?

src/devices/src/virtio/console/inorder_handler.rs
```diff
@@ -0,0 +1,97 @@
...
+                break;
+            }
+            self.receiveq.disable_notification()?;
+
+            while let Some(mut chain) = self.receiveq.iter()?.next() {
+                let used_len = self.console.process_receiveq_chain(&mut chain);
```
lauralt, line 76: In case we weren't able to write any byte to the memory buffer, we might want to go back one position into the queue. I think this is how CH and crosvm are handling this situation.

src/devices/src/virtio/console/inorder_handler.rs
```diff
@@ -0,0 +1,97 @@
...
+
+                self.receiveq.add_used(chain.head_index(), used_len as u32)?;
+                notify = true;
```
lauralt, line 79: Shouldn't we mark notify as true only if we are sure we wrote at least one byte to memory?

src/devices/src/virtio/console/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+    // pub allow_resize: bool,
+    // pub allow_multiport: bool,
+    // pub allow_emerg_write: bool,
+}
+
+impl ConsoleArgs {
+    // Generate device features based on the configuration options.
+    pub fn device_features(&self) -> u64 {
+        
+        let features = 1 << VIRTIO_F_VERSION_1;
```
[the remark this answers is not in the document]
RaduNiculae: I think we could enable VIRTIO_F_IN_ORDER, though the console already handles the descriptors in order. I do not know if we need the event suppression of the VIRTIO_F_RING_EVENT_IDX since we want to be notified every time.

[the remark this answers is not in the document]
lauralt: Synced in private on this. Resolving it.

src/vmm/src/config/mod.rs
```diff
@@ -258,6 +264,32 @@ impl TryFrom<&str> for BlockConfig {
...
+}
+
+impl TryFrom<&str> for ConsoleConfig {
+    type Error = ConversionError;
+
+    fn try_from(console_cfg_str: &str) -> Result<Self, Self::Error> {
+        // Supported options: `type=String`
+        let mut arg_parser = CfgArgParser::new(console_cfg_str);
+
+        let console_type = arg_parser
```
lauralt, line 281: I think we can also define an enum, let's say `ConsoleType`, for the possible values for `type` and if the string is not one of the expected values, throw an error. So ConsoleConfig will no longer store a `String`, but a `ConsoleType` object.

lauralt: Then we should also add tests for invalid input, example `--console type=blah`.

src/devices/src/virtio/console/inorder_handler.rs
```diff
@@ -0,0 +1,112 @@
...
+                        self.receiveq.state.go_to_previous_position();
+                        return Err(Error::Console(e));
+                    }
+                };
+                if used_len == 0 {
+                    self.receiveq.state.go_to_previous_position();
+                    break;
+                }
+                self.receiveq
+                    .add_used(chain.head_index(), used_len as u32)?;
```
lauralt, line 97: I'm not sure about the conversion here from usize to u32. Can we actually update `process_receiveq_chain` to return an u32 instead? From a quick look, it seemed to me that we can.

src/devices/src/virtio/console/inorder_handler.rs
```diff
@@ -0,0 +1,112 @@
...
+                    Ok(used_len) => used_len,
+                    Err(e) => {
+                        self.receiveq.state.go_to_previous_position();
+                        return Err(Error::Console(e));
+                    }
+                };
+                if used_len == 0 {
+                    self.receiveq.state.go_to_previous_position();
+                    break;
+                }
```
lauralt, line 95: We have to think a bit more about this. I think `process_receiveq_chain` can fail even if there were some bytes written to the chain, in which case we shouldn't go back one position in the available ring.

lauralt (commented):
Left a few other comments. Will finish the review early next week.

src/devices/src/virtio/console/inorder_handler.rs
```diff
@@ -0,0 +1,112 @@
...
+            self.transmitq.disable_notification()?;
+
+            while let Some(mut chain) = self
+                .transmitq
+                .state
+                .pop_descriptor_chain(self.transmitq.mem.memory())
+            {
+                self.console.process_transmitq_chain(&mut chain)?;
+
+                self.transmitq.add_used(chain.head_index(), 0)?;
```
lauralt, line 59: I think it would be better to check if the driver needs a notification each time we add an entry in the used ring (like we do for the other devices in vmm-ref). There's a comment about that [here](https://github.com/rust-vmm/vm-virtio/blob/main/crates/virtio-queue/src/state.rs#L416) and documented [here](https://docs.oasis-open.org/virtio/virtio/v1.1/csprd01/virtio-v1.1-csprd01.html#x1-400007).

src/devices/src/virtio/console/inorder_handler.rs
```diff
@@ -0,0 +1,112 @@
...
+
+            while let Some(mut chain) = self
+                .receiveq
+                .state
+                .pop_descriptor_chain(self.receiveq.mem.memory())
+            {
+                let used_len = match self.console.process_receiveq_chain(&mut chain) {
+                    Ok(used_len) => used_len,
+                    Err(e) => {
+                        self.receiveq.state.go_to_previous_position();
```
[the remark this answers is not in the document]
lauralt: For invalid descriptor chains, we usually add a new entry in the used ring with len = 0, but from what I see there are device implementations that don't add an entry in the used ring for that. I would be in favor for adding one.

src/api/src/lib.rs
```diff
@@ -50,6 +50,13 @@ impl Cli {
                     .required(false)
                     .takes_value(true)
                     .help("Block device configuration. \n\tFormat: \"path=<string>\"")
+            )
+            .arg(
+                Arg::with_name("console")
+                    .long("console")
+                    .required(false)
+                    .takes_value(true)
+                    .help("Console configuration. \n\tFormat: \"type=<string>\"")
```
andreeaflorescu, line 59: What are the possible types here? For the other ones it is self explanatory what the value of the configuration is (i.e. a path, kernel_load_addr, etc), but for this one it is not. We should write the possible values in the help here, and also when we are parsing it, we should check that the value is one of the supported ones.

src/api/src/lib.rs
```diff
@@ -236,6 +244,7 @@ mod tests {
                 vcpu_config: VcpuConfig { num: 1 },
                 block_config: None,
                 net_config: None,
+                console_config: None
```
andreeaflorescu, line 247: We should update the test to have a console_config that's being parsed as well. Best would be to have it with all possible values.

src/devices/src/virtio/console/mod.rs
```diff
@@ -0,0 +1,23 @@
...
+pub const CONSOLE_DEVICE_ID: u32 = 3;
+
+#[derive(Debug)]
+pub enum Error {
+    Virtio(crate::virtio::Error),
+    Console(virtio_console::console::Error),
+}
+pub type Result<T> = std::result::Result<T, Error>;
+
+pub fn device_features() -> u64 {
```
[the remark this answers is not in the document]
andreeaflorescu: Sounds good to have it in the constructor for now.

src/devices/src/virtio/console/queue_handler.rs
```diff
@@ -0,0 +1,116 @@
...
+use vmm_sys_util::eventfd::EventFd;
+
+use crate::virtio::console::inorder_handler::InOrderQueueHandler;
+use crate::virtio::SingleFdSignalQueue;
+
+const STDIN_EVENT: u32 = 0;
+const TRANSMITQ_EVENT: u32 = 1;
+const RECEIVEQ_EVENT: u32 = 2;
+
+const STDIN_BUFFER_SIZE: usize = 1024;
```
andreeaflorescu, line 16: Where is this taken from?

src/devices/src/virtio/console/queue_handler.rs
```diff
@@ -0,0 +1,116 @@
...
+                        error!("Receiveq processing failed: {:?}", e);
+                    }
+                }
+            }
+            TRANSMITQ_EVENT => {
+                if let Err(e) = self.transmitqfd.read() {
+                    error!("Could not read transmitq event fd: {:?}", e);
+                } else if let Err(e) = self.inner.process_transmitq() {
+                    error!("Transmitq processing failed: {:?}", e);
+                }
```
andreeaflorescu, line 80: Can you try to use `and_then` to reduce some of these lines of code?

src/devices/src/virtio/console/queue_handler.rs
RaduNiculae: This is just a random value. To figure out a good one we should run some tests. I tried with some python tests but I could not measure the time it takes to send data to the driver accurately.

PR #219, opened 2022-06-21
Adding the system allocator POC

### LINKED ISSUE #132: Import system allocator (once it becomes available)
Adapt the code in vmm-reference to use a system allocator for interrupts and MMIO ranges instead of having them statically assigned.

ramyak-mehra:
closes #132

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/arch/src/lib.rs
+++ b/src/arch/src/lib.rs
@@ -63,20 +63,20 @@ impl From<GuestMemoryError> for Error {
 pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Default)]
-pub struct FdtBuilder<'a> {
-    cmdline: Option<&'a str>,
+pub struct FdtBuilder {
+    cmdline: Option<String>,
     mem_size: Option<u64>,
     num_vcpus: Option<u32>,
     serial_console: Option<(u64, u64)>,
     rtc: Option<(u64, u64)>,
 }
 
-impl<'a> FdtBuilder<'a> {
+impl FdtBuilder {
     pub fn new() -> Self {
         FdtBuilder::default()
     }
 
-    pub fn with_cmdline(&mut self, cmdline: &'a str) -> &mut Self {
+    pub fn with_cmdline(&mut self, cmdline: String) -> &mut Self {
         self.cmdline = Some(cmdline);
         self
     }
@@ -114,6 +114,7 @@ impl<'a> FdtBuilder<'a> {
 
         let cmdline = self
             .cmdline
+            .as_ref()
             .ok_or_else(|| Error::MissingRequiredConfig("cmdline".to_owned()))?;
         create_chosen_node(&mut fdt, cmdline)?;
 
@@ -325,7 +326,7 @@ mod tests {
     #[test]
     fn test_create_fdt() {
         let fdt_ok = FdtBuilder::new()
-            .with_cmdline("reboot=t panic=1 pci=off")
+            .with_cmdline(String::from("reboot=t panic=1 pci=off"))
             .with_num_vcpus(8)
             .with_mem_size(4096)
             .with_serial_console(0x40000000, 0x1000)
@@ -342,7 +343,7 @@ mod tests {
         assert!(fdt_no_cmdline.is_err());
 
         let fdt_no_num_vcpus = FdtBuilder::new()
-            .with_cmdline("reboot=t panic=1 pci=off")
+            .with_cmdline(String::from("reboot=t panic=1 pci=off"))
             .with_mem_size(4096)
             .with_serial_console(0x40000000, 0x1000)
             .with_rtc(0x40001000, 0x1000)
@@ -350,7 +351,7 @@ mod tests {
         assert!(fdt_no_num_vcpus.is_err());
 
         let fdt_no_mem_size = FdtBuilder::new()
-            .with_cmdline("reboot=t panic=1 pci=off")
+            .with_cmdline(String::from("reboot=t panic=1 pci=off"))
             .with_num_vcpus(8)
             .with_serial_console(0x40000000, 0x1000)
             .with_rtc(0x40001000, 0x1000)
--- a/src/vmm/src/lib.rs
+++ b/src/vmm/src/lib.rs
@@ -208,6 +208,8 @@ pub struct Vmm {
     // TODO-continued: after the other resources are created.
     #[cfg(target_arch = "aarch64")]
     num_vcpus: u64,
+    #[cfg(target_arch = "aarch64")]
+    fdt_builder: FdtBuilder,
 }
 
 // The `VmmExitHandler` is used as the mechanism for exiting from the event manager loop.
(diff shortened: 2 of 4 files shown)


src/vmm/src/system_allocator.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{MemoryConfig, MemoryError};
+use vm_allocator::{AddressAllocator, AllocPolicy, IdAllocator, RangeInclusive};
+use vm_device::bus::{BusRange, MmioAddress, MmioRange};
+pub(crate) struct SystemAllocator {
+    address: AddressAllocator,
+}
+pub type Result<T> = std::result::Result<T, crate::Error>;
+
+impl SystemAllocator {
```
andreeaflorescu, line 9: Why are you implementing a wrapper over the `AddressAllocator`? Do you think we will be using more than one allocator? If we don't use another one, it might be a good option to just use `AddressAllocator` directly so that we don't need the wrapper methods.

andreeaflorescu: This is an early optimization, we don't know how the code will look like at that point in time, so I think it would be better instead to just use the AddressAllocator directly.

Otherwise, the integration is going in the good direction, can you do the same for the serial console, and other devices?

src/vmm/src/lib.rs
```diff
@@ -103,12 +108,18 @@ pub const DEFAULT_KERNEL_CMDLINE: &str = "panic=1 pci=off";
...
+pub const DEFAULT_ADDRESSS_ALIGNEMNT: u64 = 4;
+///Default allocation policy for address allocator.
+pub const DEFAULT_ALLOC_POLICY: AllocPolicy = AllocPolicy::FirstMatch;
 
 /// VMM memory related errors.
 #[derive(Debug)]
 pub enum MemoryError {
     /// Not enough memory slots.
     NotEnoughMemorySlots,
+    ///AddressAllocatorError
```
gsserge, line 121:
```suggestion
    /// AddressAllocatorError
```
A space is missing here and in some of the comments below.

src/vmm/src/lib.rs
```diff
@@ -361,9 +380,19 @@ impl Vmm {
                 (GuestAddress(MMIO_GAP_END), remaining),
             ],
         }
+
         #[cfg(target_arch = "aarch64")]
         vec![(GuestAddress(AARCH64_PHYS_MEM_START), mem_size)]
     }
+    fn create_address_allocator(memory_config: &MemoryConfig) -> Result<AddressAllocator> {
```
gsserge, line 387: Please add a new line.

src/vmm/src/lib.rs
```diff
@@ -515,7 +549,11 @@ impl Vmm {
     #[cfg(target_arch = "aarch64")]
     fn add_rtc_device(&mut self) {
         let rtc = Arc::new(Mutex::new(RtcWrapper(Rtc::new())));
-        let range = MmioRange::new(MmioAddress(AARCH64_MMIO_BASE + 0x1000), 0x1000).unwrap();
+        let range = self
+            .address_allocator
+            .allocate(0x1000, DEFAULT_ADDRESSS_ALIGNEMNT, DEFAULT_ALLOC_POLICY)
+            .unwrap();
```
gsserge, line 555: Can we use `?` instead of `unwrap` similar to `fn add_block_device()`?

src/vmm/src/lib.rs
```diff
@@ -647,7 +693,10 @@ impl Vmm {
         Ok(())
     }
 }
-
+fn get_mmio_from_range(range: RangeInclusive) -> MmioRange {
+    //Address allocator will make sure we have the address available.
+    MmioRange::new(MmioAddress(range.start()), range.len()).unwrap()
```
gsserge, line 698: Are we sure this cannot fail?

src/vmm/src/lib.rs
```diff
@@ -647,7 +693,10 @@ impl Vmm {
...
+}
```
gsserge, line 699: Please add a new line.

src/vmm/src/lib.rs
```diff
@@ -1093,4 +1128,98 @@ mod tests {
             assert!(fdt.write_to_mem(&vmm.guest_memory, fdt_offset).is_err());
         }
     }
+    #[test]
+    fn test_address_alloc() {
```
gsserge, line 1132: To me this looks like a set of unit tests for the allocator itself, not the vmm integration testing. Therefore I do not think this belongs here; these kind of tests belong to the allocator crate itself.

andreeaflorescu, line 1132: Let's re-write this test to avoid duplication. To do that, we need to identify what are the things that are different on different platforms. Once we do that, we need to find a common name for the things that are different, and then just initialize them with separate values depending on the platform. We can also make some simplifications, so that we don't need to define a lot of variables.

For example, let's say we have a `start_addr` from which addresses are valid. Then, we can define an `invalid_addr` as being `start_addr - 100`. While `start_addr` needs to be defined once for each platform, the `invalid_addr` will just be defined once for both platforms.

andreeaflorescu (commented):
I would propose to squash all commits into one because and avoid the clippy fixes as separate commits.

src/vmm/src/lib.rs
andreeaflorescu: I would like us to keep parts of this tests, specifically the negative ones where we try to allocate something that does not fit in the MMIO range defined for each platform. That one would test that we initialized the AddressAllocator properly, but it's also a bit of a simplistic test that I was considering to be a "safety net" testing that the allocator behaves as we expect it to.

src/vmm/src/lib.rs
ramyak-mehra: since we are passing the range, which we get from the address allocator, address allocator takes care of making sure there is always address available.

ramyak-mehra:
> I would propose to squash all commits into one because and avoid the clippy fixes as separate commits.

I would squash and commit the last commit after everything has been been approved

andreeaflorescu (changes requested):
As we discussed online, the FDT configuration is currently not correct because the address of the serial console and the RTC devices are currently hard coded.

We should define an `FDTWriter` in the `Vmm` and that initialize the FDT entries as we are allocating the ranges.

src/vmm/src/lib.rs
```diff
@@ -1093,4 +1147,48 @@ mod tests {
...
+        let memory_config = MemoryConfig {
+            size_mib: MEM_SIZE_MIB,
+        };
+        #[cfg(target_arch = "x86_64")]
+        let start_addr = MMIO_GAP_START;
+        #[cfg(target_arch = "aarch64")]
+        let start_addr = AARCH64_MMIO_BASE;
+        let mut address_alloc = Vmm::create_address_allocator(&memory_config).unwrap();
+        // Trying to allocate at an address before the base address.
```
andreeaflorescu, line 1160: nit: Can you add a white line between each test case? The code looks a bit crowded.

src/arch/src/lib.rs
```diff
@@ -63,20 +63,20 @@ impl From<GuestMemoryError> for Error {
 pub type Result<T> = std::result::Result<T, Error>;
 
 #[derive(Default)]
-pub struct FdtBuilder<'a> {
-    cmdline: Option<&'a str>,
+pub struct FdtBuilder {
+    cmdline: Option<String>,
```
andreeaflorescu, line 67: When doing the squash, please leave this change in a separate commit as it is only mildly related to the rest of the changes. You should add an explanation in the commit message why was this change added.

andreeaflorescu (commented):
LGTM. Once you squash the commits I'll approve this PR.

andreeaflorescu (approved):

gsserge (approved):

PR #223, opened 2022-07-12
Adding virtio support for arm

### LINKED ISSUE #210: Add support for virtio block for aarch64
Configure the virtio block device on arm as well. For this, we need to update the FDT builder to support adding virtio block devices, and update the unit tests such that they work on arm as well.

The definition of done for this task is to:
- make it possible to configure one virtio block
- add tests to validate that this configuration work (we should be able to reuse the existing tests from x86_64)

Out of scope:
- the end to end integration test is out of scope because it requires extensive changes to the S3 bucket & the way we're building the test resources. Any changes that are required once we have a test image should be part of that PR instead.

ramyak-mehra:
Closes #210

### Summary of the PR

### Requirements

requirements:

  description line.
  test.

### DIFF
--- a/src/arch/src/lib.rs
+++ b/src/arch/src/lib.rs
@@ -62,6 +62,13 @@ impl From<GuestMemoryError> for Error {
 
 pub type Result<T> = std::result::Result<T, Error>;
 
+/// It contains info about the virtio device for fdt.
+struct DeviceInfo {
+    addr: u64,
+    size: u64,
+    irq: u32,
+}
+
 #[derive(Default)]
 pub struct FdtBuilder {
     cmdline: Option<String>,
@@ -69,6 +76,7 @@ pub struct FdtBuilder {
     num_vcpus: Option<u32>,
     serial_console: Option<(u64, u64)>,
     rtc: Option<(u64, u64)>,
+    virtio_devices: Vec<DeviceInfo>,
 }
 
 impl FdtBuilder {
@@ -99,6 +107,15 @@ impl FdtBuilder {
     pub fn with_rtc(&mut self, addr: u64, size: u64) -> &mut Self {
         self.rtc = Some((addr, size));
         self
+    }
+
+    pub fn add_virtio_device(&mut self, addr: u64, size: u64, irq: u32) -> &mut Self {
+        self.virtio_devices.push(DeviceInfo { addr, size, irq });
+        self
+    }
+
+    pub fn virtio_device_len(&self) -> usize {
+        self.virtio_devices.len()
     }
 
     pub fn create_fdt(&self) -> Result<Fdt> {
@@ -136,7 +153,9 @@ impl FdtBuilder {
         create_timer_node(&mut fdt, num_vcpus)?;
         create_psci_node(&mut fdt)?;
         create_pmu_node(&mut fdt, num_vcpus)?;
-
+        for info in &self.virtio_devices {
+            create_virtio_node(&mut fdt, info.addr, info.size, info.irq)?;
+        }
         fdt.end_node(root_node)?;
 
         Ok(Fdt {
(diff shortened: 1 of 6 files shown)


src/arch/src/lib.rs
```diff
@@ -100,7 +102,13 @@ impl FdtBuilder {
         self.rtc = Some((addr, size));
         self
     }
-
+    pub fn add_virito_device(&mut self, addr: u64, size: u64, irq: u32) -> &mut Self {
```
andreeaflorescu, line 105: typo: virito instead of virtio.

src/arch/src/lib.rs
```diff
@@ -100,7 +102,13 @@ impl FdtBuilder {
...
+        self.virtio_devices.push((addr, size, irq));
+        self
+    }
```
andreeaflorescu, line 108: Please add a blank line after each function definition. This is needed for all function you're defining.

src/arch/src/lib.rs
```diff
@@ -319,10 +329,29 @@ fn create_rtc_node(fdt: &mut FdtWriter, rtc_addr: u64, size: u64) -> Result<()>
     Ok(())
 }
 
+fn create_virito_node(fdt: &mut FdtWriter, addr: u64, size: u64, irq: u32) -> Result<()> {
```
andreeaflorescu, line 332: typo: virito -> virtio.

Same typo on line 333.

src/devices/src/virtio/mod.rs
```diff
@@ -331,11 +336,24 @@ pub(crate) mod tests {
                 kernel_cmdline: &mut self.kernel_cmdline,
             }
         }
+        #[cfg(target_arch = "aarch64")]
+        // Support only GIC_V3.
```
andreeaflorescu, line 341: Let's add more comments here: adding a dummy GIC V3 so that we can register the IRQ fds and thus test the virtio functionality in arm as well.

src/vmm/src/lib.rs
```diff
@@ -564,7 +564,7 @@ impl Vmm {
             DEFAULT_ALLOC_POLICY,
         )?;
         self.fdt_builder.with_rtc(range.start(), range.len());
-        let range = mmio_from_range(range);
+        let range = mmio_from_range(&range);
```
andreeaflorescu, line 567: We should have this change in a separate commit. If that's too much hassle, we can keep it in this commit but we need to be explicit in the commit message to why this change is needed.

andreeaflorescu (commented):
Can you add more details in the commit message? I am expecting to see stuff like:
- a description of what is needed for adding these devices on arm (i.e. the fdt)
- what other changes were required for this to be possible (i.e. the updates on the Mock interface, why were they needed? because otherwise you cannot register_irqfd)
- any changes that you did as part of this commit that are not directly related to the support (i.e. the `mmio_from_range` receives a reference now).

This subtleties are not evident from reading the code and there is no other way of understanding why this commit introduces the changes that it does.

src/arch/src/lib.rs
```diff
@@ -319,10 +338,35 @@ fn create_rtc_node(fdt: &mut FdtWriter, rtc_addr: u64, size: u64) -> Result<()>
...
     use crate::FdtBuilder;
 
+    #[test]
+    fn test_adding_virtio() {
+        let mut fdt = FdtBuilder::new();
+        let fdt = fdt.add_virtio_device(0x1000, 1000, 5);
+        assert_eq!(fdt.virtio_devices.len(), 1);
+    }
+    #[test]
+    fn test_virtio_device_len() {
```
andreeaflorescu, line 365: This test and the one above seem to be doing the same thing. Can you delete one of them?

gsserge (approved):

andreeaflorescu (approved):

gsserge (approved):

PR #225, opened 2022-07-19
Made relevant changes to generate resources for aarch64.

ramyak-mehra:
Added support to generate deb based pe linux image.
Added support to genreate .ext4 disk to boot from.
Added v5 files to resoure_manifest.json.

### Summary of the PR

### Requirements

requirements:

  description line.
  test.


.buildkite/build_resources.py
```diff
@@ -27,6 +36,9 @@
         for v in json.loads(content):
             if v['filename'].startswith('resources/'):
                 pipeline['steps'][0]['skip'] = False
+                pipeline['steps'][0]['build']['message'] = os.environ.get('BUILDKITE_MESSAGE')
```
andreeaflorescu, line 39: We should add a comment here about why is it not working to use the definition from the `pipeline` dictionary.

Can you also update the commit message to say "Fixed the build resources pipeline" in the summary, and then add details in the description as well with what was the problem (i.e. it was always running on main).

tests/tools/s3/resource_manifest.json
```diff
@@ -255,5 +255,30 @@
...
+    {
+      "resource_type": "disk",
+      "resource_name": "ubuntu-focal-rootfs.ext4",
+      "relative_path": "disk/",
+      "distro": "Ubuntu",
+      "compatible_kernels": "v5/kernel/pe-focal|v5/kernel/pe-focal-halt"
+    }
   ]
-}
+}
```
andreeaflorescu, line 284: Can you keep the new line as it was before?

andreeaflorescu (commented):
Can you also do a `git pull --rebase upstream main` when you address the other comments so we can get rid of the commit in the middle?

resources/make_test_resources.sh
```diff
@@ -56,4 +56,14 @@ elif [[ $arch = "aarch64" ]]; then
...
     mv $TMP_KERNEL_DIR/linux-5.4.81/initramfs.cpio $DEST_INITRD_DIR/aarch64-initramfs.cpio
+
+    # making debian based image
+    ./kernel/make_kernel_image_deb.sh -f pe -k pe-focal -w $TMP_DEB_DIR -j $NPROC
+    ./kernel/make_kernel_image_deb.sh -f pe -k pe-focal -w $TMP_DEB_DIR -j $NPROC -h
+
+    #  move them to correct folder
+    mv $TMP_DEB_DIR/linux-5.4.81/pe-focal $TMP_DEB_DIR/linux-5.4.81/pe-focal-halt $DEST_KERNEL_DIR
+
+    ./disk/make_rootfs.sh -d $TMP_DEB_DIR/linux-5.4.81/deb/ -w $DEST_DISK_DIR -o ubuntu-focal-rootfs.ext4
```
andreeaflorescu, line 67: This needs to be renamed to ubuntu-focal-rootfs-aarch64.ext4 because otherwise it will have the same name as the x86_64 one and we cannot upload them to S3.

.buildkite/build_resources.py
```diff
@@ -27,6 +36,11 @@
         for v in json.loads(content):
             if v['filename'].startswith('resources/'):
                 pipeline['steps'][0]['skip'] = False
+                # We are doing this to override variable to point to the branch/PR
+                # from where the changes were pushed.
```
andreeaflorescu, line 40: We should also say that we cannot use the variable names directly because `yaml.dump` does not escape the quotes properly.

tests/test_run_reference_vmm.py
```diff
@@ -61,8 +61,10 @@ def default_ubuntu_elf():
 
 def default_disk():
     return s3_download(
-        resource_type="disk",
-        resource_name="ubuntu-focal-rootfs.ext4",
+        resource_type="disk",        
+        resource_tags={
+            "platform_type":"x86_64"
```
andreeaflorescu, line 66: Here we should just read the platform on which we're running and use that one instead. We can also do this in a subsequent PR in which we also update the tests. The goal would be to reuse as many functions as possible on both platforms.

ramyak-mehra: I was planning to do it when we port of the tests from x86.

andreeaflorescu: Sounds good.

andreeaflorescu:
We also need to update the `CONTAINER_VERSION` file to 15 as this is the version that we're using for building the resources.

resources/kernel/make_kernel_image_deb.sh
```diff
@@ -18,7 +19,12 @@ source "$TEST_RESOURCE_DIR/make_kernel.sh"
 
 KERNEL_VERSION="5.4.81"
 
-KERNEL_CFG="microvm-kernel-5.4-x86_64.config"
+if [[ $arch = "x86_64" ]]; then
+	KERNEL_CFG="microvm-kernel-5.4-x86_64.config"
```
lauralt, line 23: I think you can construct KERNEL_CFG based on the arch value, instead of having this if else.

resources/kernel/make_kernel_image_deb.sh
```diff
@@ -103,11 +109,19 @@ if [ -n "$MAKEDEB" ]; then
     cp "$kernel_dir"/../*.deb "$kernel_dir/deb"
 else
     echo "Downloading deb packages..."
+    if [[ $arch = "x86_64" ]]; then
```
andreeaflorescu, line 112: This if/elseif is missing formatting. The code inside the blocks should be aligned with one tab more to the right.

tests/tools/s3/resource_manifest.json
```diff
@@ -255,5 +255,122 @@
...
+      "relative_path": "kernel/",
+      "image_format": "pe",
+      "halt_after_boot": false,
+      "with_disk": true,
+      "platform_type": "aarch64",
+      "distro": "Ubuntu"
+    },
+    {
+      "resource_type": "kernel",
+      "resource_name": "pe-focal-halt",
```
lauralt, line 346: Do we need this one? I see that on x86, we don't keep the focal resources that are halting.

ramyak-mehra: inside the make_test_resources.sh file it does make focal resources that are halting.

lauralt: Yeah, I know that but I think we didn't also upload them to S3. I was asking if we need them somewhere in the tests.

andreeaflorescu: We might need them in the future. It makes the upload process easier if all the files that are generated are also uploaded. Otherwise we need to update the script as to not generate them anymore. The aarch64 does not currently run all tests that run on x86_64. To stay on the safe side I would say to keep the pe-focal-halt image as well.

resources/kernel/make_kernel_image_deb.sh
ramyak-mehra: there are some changes that was required in the default config.
earlier also a pre made config file was present

tests/tools/s3/resource_manifest.json
ramyak-mehra: should I remove them in a different commit, stating the reason we don't actually need them?
or in a previous commit itself

lauralt: Separate commit sounds good! It can be placed wherever, I don't have a strong preference about it.

andreeaflorescu (approved):

PR #227, opened 2022-08-05
Irq allocator

ramyak-mehra:
### Summary of the PR

### Requirements

requirements:

  description line.
  test.


src/api/src/lib.rs
```diff
@@ -31,6 +31,12 @@ impl Cli {
                     .takes_value(true)
                     .help("vCPU configuration.\n\tFormat: \"num=<u8>\""),
             )
+            .arg(
```
andreeaflorescu, line 34: I think for now it's better to not export it at the API level as well because we don't have a customer need for it yet. Going forward we want to have a VMM library, and we can export it in the VMM interface, but for the API we should just keep it simple.

src/vm-vcpu/src/vm.rs
```diff
@@ -229,7 +231,11 @@ impl<EH: 'static + ExitHandler + Send> KvmVm<EH> {
         let mut vm = Self::create_vm(kvm, vm_config, exit_handler, guest_memory)?;
 
         #[cfg(target_arch = "x86_64")]
-        MpTable::new(vm.config.num_vcpus)?.write(guest_memory)?;
+        {
+            let mp_table = MpTable::new(vm.config.num_vcpus)?;
+            mp_table.write(guest_memory)?;
+            self.irq_num = Some(mp_table.irq_num())
```
andreeaflorescu, line 237: This is no longer needed, should be deleted.

src/vm-vcpu/src/vm.rs
```diff
@@ -229,7 +231,11 @@ impl<EH: 'static + ExitHandler + Send> KvmVm<EH> {
         let mut vm = Self::create_vm(kvm, vm_config, exit_handler, guest_memory)?;
 
         #[cfg(target_arch = "x86_64")]
-        MpTable::new(vm.config.num_vcpus)?.write(guest_memory)?;
+        {
+            let mp_table = MpTable::new(vm.config.num_vcpus)?;
```
andreeaflorescu, line 235: I think it makes more sense to send the IRQ as a parameter for the MpTable similarly to how we are setting the vcpu number. This is also what we do on aarch64 when passing the IRQ number to the GIC config.

src/vm-vcpu/src/vm.rs
```diff
@@ -652,7 +664,6 @@ mod tests {
             ranges.push((GuestAddress(i * 100), 100))
         }
         let guest_memory = GuestMemoryMmap::from_ranges(&ranges).unwrap();
-
```
andreeaflorescu, line 655: Can you leave the whitespaces as they were? They are separating configuration and initialization and is making the code easier to read.

src/vmm/src/irq_allocator.rs
```diff
@@ -0,0 +1,84 @@
...
+// to cater more complex scenario.
+#[derive(Debug)]
+pub struct IrqAllocator {
+    // Tracks the last allocated irq
+    last_used_irq: u32,
+    max: u32,
+}
+
+impl IrqAllocator {
+    pub fn new(last_used_irq: u32, max: u32) -> Result<Self> {
```
andreeaflorescu, line 18: For the interface it doesn't make sense for this parameter to be called `last_used_irq` because the caller typically wants to initialize a range of available IRQs. Names such as `first_irq` and `last_irq` are more appropriate for this initialization. It is also more clear that with these we mean the actual value, and not a size. From that point of view `max` is an ambiguous name because it's unclear whether it's the irq, or the size of the interval.

You can keep the name as `last_used_irq` in the field name because that's the intention with which we're going to use it.

src/vmm/src/irq_allocator.rs
```diff
@@ -0,0 +1,84 @@
...
+        if last_used_irq >= max {
+            return Err(Error::InvalidValue);
+        }
+        Ok(IrqAllocator { last_used_irq, max })
+    }
+
+    pub fn next_irq(&mut self) -> Result<u32> {
+        match self.last_used_irq.checked_add(1) {
```
andreeaflorescu, line 26: You can write this in a more Rust-y way by making use of `ok_or` which converts the checked_add result from an Option to an Error.

src/vmm/src/irq_allocator.rs
```diff
@@ -0,0 +1,84 @@
...
+    use super::{Error, IrqAllocator};
+    #[test]
+    fn test_new() {
+        let irq_alloc = IrqAllocator::new(4, 10).unwrap();
+        assert_eq!(irq_alloc.last_used_irq, 4);
+        assert_eq!(irq_alloc.max, 10);
+        let irq_alloc = IrqAllocator::new(4, 4).unwrap_err();
+        assert_eq!(irq_alloc, Error::InvalidValue);
+        let irq_alloc = IrqAllocator::new(4, 3).unwrap_err();
+        assert_eq!(irq_alloc, Error::InvalidValue);
```
andreeaflorescu, line 72: You should add a test where the max is u32::MAX and check for overflow.

src/vmm/src/lib.rs
```diff
@@ -299,10 +309,13 @@ impl TryFrom<VMMConfig> for Vmm {
         #[cfg(target_arch = "aarch64")]
         let fdt_builder = FdtBuilder::new();
 
+        let irq_allocator = IrqAllocator::new(4, vm.max_irq())?;
```
andreeaflorescu, line 312: We should have a define for `4` instead and use it when we initialize the console on both platforms.

src/vmm/src/irq_allocator.rs
ramyak-mehra: I was thinking of that, but the check that `last_used_irq >= max` prevents me from adding it at the first place.

src/vm-vcpu/src/vm.rs
ramyak-mehra: Yes, I was going first for finalising the api for aarch64 of passing down the irq, will fix x86 next.

src/vmm/src/irq_allocator.rs
ramyak-mehra: I changed max to `last_irq` but for `last_used_irq` I think because the user can never get lets say 4 so it's like what he actually last allocated.

src/vmm/src/irq_allocator.rs
andreeaflorescu: That's good, but we should have a test for checking the extreme values as well so that in the future if the check is gone we still don't get an overflow. We don't need to check for the overflow error specifically, but rather that we don't receive a panic.

src/vmm/src/config/mod.rs
```diff
@@ -132,7 +139,38 @@ impl TryFrom<&str> for VcpuConfig {
...
+    pub max_irq: u32,
+}
+
+impl Default for IrqConfig {
+    fn default() -> Self {
+        IrqConfig { max_irq: MAX_IRQ.into() }
+    }
+}
 
+impl TryFrom<&str> for IrqConfig {
```
andreeaflorescu, line 155: We don't need this since we are not going to configure the irq from the command line. We should just call `VmConfig::new(&kvm, config.vcpu_config.num, MAX_IRQ)?` instead, and remove the definition of `IrqConfig`.

src/vm-vcpu/src/vm.rs
```diff
@@ -27,19 +28,26 @@ use vm_vcpu_ref::aarch64::interrupts::{self, Gic, GicConfig, GicState};
 #[cfg(target_arch = "x86_64")]
 use vm_vcpu_ref::x86_64::mptable::{self, MpTable};
 
+#[cfg(target_arch = "aarch64")]
+pub const MAX_IRQ: u32 = interrupts::MIN_NR_IRQS;
+#[cfg(target_arch="x86_64")]
+pub const MAX_IRQ:u8 = mptable::IRQ_MAX;
```
andreeaflorescu, line 34: Let's define this as u32 as well so that we can keep the code simpler for both platforms. If we want to have one as u32 and one as u8, then this type difference should be passed to the `VmConfig` structure as well, and that would needlessly complicate the code. Instead, we can define both as u32.

On x86, for now, we need the IRQ to be an u8. We can do a `try_from` when we initialize the `MpTable`, and return an error if the number is too big. I am saying for now because in the future we might choose to switch to I/O APIC and thus increase the number of available IRQs.

src/vm-vcpu/src/vm.rs
```diff
@@ -229,7 +237,10 @@ impl<EH: 'static + ExitHandler + Send> KvmVm<EH> {
         let mut vm = Self::create_vm(kvm, vm_config, exit_handler, guest_memory)?;
 
         #[cfg(target_arch = "x86_64")]
-        MpTable::new(vm.config.num_vcpus)?.write(guest_memory)?;
+        {
+            let max_irq:u8 = vm.config.max_irq.try_into().expect("max irq should be u8");
```
andreeaflorescu, line 241: We shouldn't use expect here. Let's return an error instead. We should also add a negative test here to check that when passing a IRQ num > u8::MAX the code does not panic and rather returns an error.

src/vm-vcpu-ref/src/x86_64/mptable.rs
```diff
@@ -148,17 +148,22 @@ pub struct MpTable {
...
     /// Creates a new MP Table that can hold `cpu_num`.
-    pub fn new(cpu_num: u8) -> Result<MpTable> {
+    pub fn new(cpu_num: u8, max_irq: u8) -> Result<MpTable> {
         if cpu_num > MAX_SUPPORTED_CPUS {
             return Err(Error::TooManyCpus);
         }
 
         Ok(MpTable {
-            irq_num: IRQ_MAX + 1,
+            irq_num: max_irq + 1,
```
andreeaflorescu, line 157: We should check that this addition does not overflow and return an error in case it does.

src/vm-vcpu-ref/src/x86_64/mptable.rs
```diff
@@ -148,17 +148,22 @@ pub struct MpTable {
...
             cpu_num,
         })
     }
 
+    /// Returns the number of irqs
```
andreeaflorescu, line 162: I think this is no longer needed because we are returning the IRQ from the config. Same goes for the `irq_num` function in the `Gic`.

PR #232, opened 2022-09-07
Adding python test for net device

### LINKED ISSUE #224: Add support for virtio net for aarch64
Configure the virtio net device on arm as well. For this, we need to update the FDT builder to support adding virtio block devices, and update the unit tests such that they work on arm as well.

The definition of done for this task is to:
- make it possible to configure one virtio net
- add tests to validate that this configuration work (we should be able to reuse the existing tests from x86_64)

Out of scope:
- the end to end integration test is out of scope because it requires extensive changes to the S3 bucket & the way we're building the test resources. Any changes that are required once we have a test image should be part of that PR instead. Handled in #218

ramyak-mehra:
closes #224

### Summary of the PR

### Requirements

requirements:

- [ ] All commits in this PR are signed (with `git commit -s`), and the commit
  description line.
  test.
- [ ] Any newly added `unsafe` code is properly documented.


ramyak-mehra:
I wanted to get a review on this, I know this is an old PR, I will do a rebase and push the with latest commits once everything seems fine from the code perspective

tests/test_run_reference_vmm.py
```diff
@@ -222,7 +226,24 @@ def run_cmd_inside_vm(cmd, vmm_process, prompt, timeout=5):
 
        Note: `cmd` and `prompt` should be a `bytes` object and not `str`.
     """
-    cmd = cmd.strip() + b'\r\n'
+    cmd = cmd.strip()
+    
+    # The max capacity of the buffer for serial console is 64.(https://github.com/rust-vmm/vm-superio/blob/282bae92878936086606715412494fa81151efba/crates/vm-superio/src/serial.rs#L34)
+    # So we split the command at 63 bytes. 1 extra is reserved for \r at the end.
+    # empty string is of size 33. sys.getsizeof(b'') = 33.
```
andreeaflorescu, line 233: Hmmm, this is rather strange. Why would the empty string require 33 bytes? Do you know why that is the case?

tests/test_run_reference_vmm.py
```diff
@@ -222,7 +226,24 @@ def run_cmd_inside_vm(cmd, vmm_process, prompt, timeout=5):
...
+    # 63-33 = 30. So we add at max 30 [0-29] chars inside the command buffer.
+    break_at = 30
+    size = sys.getsizeof(cmd)
+    while size >= 64:
+        cmd_i = cmd[:break_at]
+
+        vmm_process.stdin.write(cmd_i)
+        vmm_process.stdin.flush()
+        time.sleep(1)
```
andreeaflorescu, line 242: Why do we need the sleep?

tests/test_run_reference_vmm.py
```diff
@@ -423,3 +444,80 @@ def test_reference_vmm_mem(kernel):
...
+    
+
+def setup_guest(vmm_process , prompt , guest_ip_with_mask="172.16.0.2/24" , host_ip="172.16.0.1"):
+    """Configure the guest vm with the tap device"""
+    
+    # To clear the inital stdout which was printed while 
+    # booting the vmm.
+    expect_string(vmm_process,prompt)
+    
+    add_addr_cmd = f"ip addr add {guest_ip_with_mask} dev eth0"
```
andreeaflorescu, line 496: The name of the interface should also be sent as a parameter because it depends on the image with which you test.

tests/test_run_reference_vmm.py
```diff
@@ -423,3 +444,80 @@ def test_reference_vmm_mem(kernel):
...
+    show_interface_cmd = "ip a"
+    
+    all_cmds = [add_addr_cmd , interface_up_cmd , add_route_cmd]
+
+    for cmd in all_cmds:        
+        _ = run_cmd_inside_vm(cmd.encode() , vmm_process , prompt.encode())
+ 
+    # verifying that guest interface is setup properly
+    output = run_cmd_inside_vm(show_interface_cmd.encode() , vmm_process ,prompt.encode() , timeout=10 )    
+    output = b''.join(output).decode()
```
andreeaflorescu, line 508: Why is this needed?

andreeaflorescu (commented):
Sorry for the late review.

